PR #7, opened 2018-08-03
Initial commit for ltp lite testcase

rasibley:


veruu:
Sure, that works! Following with my comment on IRC on Friday (mentioning it here for accountability), it would be enough to include the newest current version of the tests. That should also make it easier what to include.

rasibley:
Yes I agree, I saw your ping and had planned to only include the latest version in the next PR, thanks!

PR #9, opened 2018-08-21
Fix LTP on rhel8

spbnick:
This makes the LTP test package dependencies missing on RHEL8 optional, and fixes RHEL8 support. I'm not sure if we should leave the `METADATA` target in the Makefile's at all, though.


spbnick:
Also not sure, if we should have all the dependencies we have right now.

PR #19, opened 2018-10-04
Increase the threshold for timing tests on VMs

veruu:
They often fail with "XXX slept too long". We still want to run these
test to make sure the functionality is OK, so let's try to increase the
time they can take. One of the possible causes why the old thresholds
are not enough anymore are the Spectre/Meltdown patches, since the
number of times these tests started failing increased significantly
after the mitigations were applied.


spbnick:
Looks good. Let's try that. Thank you, Veronika.

PR #21, opened 2018-10-08
Don't report timing failures on virtual systems

veruu:
Relaxing the timing thresholds doesn't seem to fix the failures. Since
multiple timing tests are failing on known good kernels, the tests
aren't suitable to run on VMs in the CI. Therefore, let's disable the
timing failure reporting while still running the tests to make sure the
functionality didn't break.


spbnick:
Looks good, thank you :)

PR #23, opened 2018-10-10
Fix kernel install on distros using BLS

veruu:
The previously added new-kernel-pkg doesn't work on these systems.
Detect the missing script and use the recommended way to install the
kernel there.

Use grubby to set the default boot entry because there is an issue with
either grub2-set-default or kernel-install which makes us unable to set
a custom kernel as the default using grub2-set-default. The workaround
is noted in the comment and can be removed after the problem is
resolved.


distribution/kpkginstall/runtest.sh
```diff
@@ -89,7 +83,15 @@ if [ ${REBOOTCOUNT} -eq 0 ]; then
       ;;
   esac
 
-  new-kernel-pkg -v --mkinitrd --dracut --depmod --make-default --host-only --install ${KVER} >>${OUTPUTFILE} 2>&1
+  if [ ! -x /sbin/new-kernel-pkg ]; then
+    kernel-install add ${KVER} /boot/vmlinuz-${KVER} >>${OUTPUTFILE} 2>&1
+    # FIXME we should use the below because grubby is getting deprecated but
+    # there's a bug in grub2-set-default or kernel-install right now
```
spbnick: Can you link to the bug in question here?

spbnick:
Thank you, Veronika!

PR #24, opened 2018-10-11
LTP: temporarily disable migrate_pages02

veruu:
See https://github.com/linux-test-project/ltp/issues/407
The test can be re-enabled after the issue is fixed and patch ported.


spbnick:
Thank you, Veronika!

PR #28, opened 2018-10-22
kpkginstall: Stop requiring KVER parameter

spbnick:
Instead of requiring the KVER parameter being passed through the
environment in distribution/kpkginstall, extract the kernel version from
the kernel image filename in the tarball.

This simplifies the interface and workings of kpet, so that it only has
to deal with a tarball, and doesn't have to know which kernel version
it's dealing with.


spbnick:
@rasibley, could you please take a look?

rasibley:
@spbnick looks good to me, I agree it makes sense to simplify if it can be extracted from the filename in the tarball.

spbnick:
Merged! Thank you, Rachel :)

PR #32, opened 2018-11-03
[pull request] xfstests: move to proper location

jencce:
Some minor fixes.


spbnick:
This looks good overall. Thank you, Xiong.

I have one question about the "xfstests: remove some RHEL unsupported testcases": is this only concerning RHEL? Would those normally work on e.g. Fedora? I.e. could they be useful for upstream kernel testing?

jencce:
Very good point @spbnick !
I tested on RHEL-7/8 for xfs and ext4, which is 4 setups. And I removed testcases that is "not run" in at least 3 out of these 4 setups.
You are right. They may be useful for Fedora or upstream kernel. We'd better keep them. I can revert that commit before you merge this PR.

spbnick:
Let's leave them on, if testing on RHEL7/8 won't break. Thank you, @jencce!

jencce:
I've done the revert.This PR can be merged if no more fix needed. Thanks!

spbnick:
Thank you, @jencce! This should be good to go. However, could you please simply remove the two last commits from this PR (which remove and then add tests), so there's less noise in the Git history? Thank you.

jencce:
Sure, done. Thanks for reviewing!

spbnick:
Thank you, @jencce, merged!

PR #34, opened 2018-11-12
xfstests: update known_issues

jencce:
Add generic/466 to known_issues. I'm still investigating this failure on F28.
Skip all tests in known_issues, then they wont report any failure.
Re-enabled TEST_PARAM_RUNTESTS, it's null by default, and we run tests in RUNTEST by default.
If we add something in this param, xfstests will run them and ignore RUNTEST file.

Thanks!

@rasibley


veruu:
@spbnick I'm not familiar with the test any more than you are, so my opinion is exactly the same :)

spbnick:
Alright, then it's up to @rasibley :grin:

rasibley:
@spbnick @veruu @jencce looks good! Tested in beaker, I'll add a comment to the internal ticket, @jencce thanks for working on this!

rasibley:
@spbnick or @veruu can you please merge ?

PR #38, opened 2018-11-16
xfstests: cleanup

jencce:
Clean up for running on Fedora.


veruu:
@spbnick @rasibley do you want to check this pull out? The changes look good to me but I'm absolutely not familiar with the test so asking for a bonus set of eyes :)

rasibley:
@jencce I noticed you were removing a lot of rhel specific code, this version of the test currently runs on rhel7, rhel8, and Fedora, just wanted to confirm that any changes you made will not prevent it from running on all releases (including RHEL) ?

jencce:
@rasibley  Sure. It will not break running on RHEL and Fedora. I have it tested. The removal is for running smoothly on Fedora without breaking running on RHEL.

rasibley:
@jencce thanks for confirming! @veruu I reviewed and don't see any reason we can't merge this, although I'm not an expert of xfstests, but it looks reasonable to me :-)

veruu:
Thanks for checking! Merging!

PR #39, opened 2018-11-16
increase timeout for LTP_TIMEOUT_MUL

rasibley:
@veruu or @spbnick  can you please review ?


rasibley:
It was decided to disable vs guessing the correct time out, will file a new PR to disable.

PR #42, opened 2018-11-16
initial commit for kvm-unit-test

rasibley:
@veruu can you please review ? This is where we will fetch kvm-unit-tests once we have it enabled in kpet.

### DIFF
--- a/vm/kvm-unit-tests/runtest.sh
+++ b/vm/kvm-unit-tests/runtest.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+. /usr/bin/rhts_environment.sh
+
+export ACCEL=kvm
+unset ARCH
+unset STANDALONE
+cpus=$(grep -c ^processor /proc/cpuinfo)
+
+# test can only run on hardware that supports virtualization 
+if egrep -q '(vmx|svm)' /proc/cpuinfo; then
+    echo "Hardware supports virtualization, proceeding" | tee -a $OUTPUTFILE
+else
+    echo "Skipping test, CPU doesn't support virtualization" | tee -a $OUTPUTFILE
+    rhts-report-result $TEST SKIP $OUTPUTFILE
+    exit
+fi
+
+# test should only run on a system with 1 or more cpus
+if [ "$cpus" > 1 ]; then
+    echo "You have sufficient CPU's to run the test" | tee -a $OUTPUTFILE
+else
+    echo "Skipping test, system requires > 1 CPU" | tee -a $OUTPUTFILE
+    rhts-report-result $TEST SKIP $OUTPUTFILE
+    exit
+fi
+
+# test is only supported on x86_64
+if [ "$(uname -i)" == "x86_64" ]; then
+    echo "Running on x86_64 arch" | tee -a $OUTPUTFILE
+else
+    echo "Skipping test, test is only supported on x86_64" | tee -a $OUTPUTFILE
+    rhts-report-result $TEST SKIP $OUTPUTFILE
+    exit
+fi
+
+# set the virt kernel parameters
+echo -e "options kvm force_emulation_prefix=1\noptions kvm enable_vmware_backdoor=1" > /etc/modprobe.d/kvm-ci.conf
+
+# reload the modules
+rmmod kvm_intel kvm_amd kvm
+modprobe kvm_intel kvm_amd kvm
+
+KVM_SYSFS=/sys/module/kvm/parameters/
+KVM_OPTIONS="enable_vmware_backdoor force_emulation_prefix"
+
+for opt in $KVM_OPTIONS; do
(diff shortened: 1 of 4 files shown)


vm/kvm-unit-tests/runtest.sh
```diff
@@ -0,0 +1,112 @@
...
+    echo "Skipping test, CPU doesn't support virtualization" | tee -a $OUTPUTFILE
+    rhts-report-result $TEST SKIP $OUTPUTFILE
+    exit
+fi
+
+# test should only run on a system with 1 or more cpus
+if [ "$cpus" > 1 ]; then
+    echo "You have sufficient CPU's to run the test" | tee -a $OUTPUTFILE
+else
+    echo "Skipping test, system requires > 1 CPU" | tee -a $OUTPUTFILE
```
veruu: Should the `rhts-report-result $TEST SKIP $OUTPUTFILE` be included in this branch too?

vm/kvm-unit-tests/runtest.sh
```diff
@@ -0,0 +1,112 @@
...
+# if running on rhel8, use python3
+if grep --quiet 8.0 /etc/redhat-release;then
+    ln -s /usr/libexec/platform-python /usr/bin/python
+fi
+
+# enable nmi_watchdog since some unit tests depend on this
+echo 0 > /proc/sys/kernel/nmi_watchdog
+
+# clone the upstream kvm-unit-tests
+git clone git://git.kernel.org/pub/scm/virt/kvm/kvm-unit-tests.git
```
veruu: I'd prefer we have a check if this command completed successfully (and abort if not). It's an external service and we can't predict any downtime. What do you think?

I also have a question purely out of curiosity since I'm not familiar with the test - is the master branch (which seems to be used) stable? Can we encounter any issues with tests being updated?

vm/kvm-unit-tests/runtest.sh
rasibley: @veruu your right, I added it

vm/kvm-unit-tests/runtest.sh
rasibley: @veruu we are picking a specific commit, will update periodically, also added some lines to abort if failed to clone/checkout.

rasibley:
@veruu can you review again ? Also just an fyi this is rhel8 only, will most likely need updates for other releases, will do this soon.

vm/kvm-unit-tests/runtest.sh
```diff
@@ -0,0 +1,118 @@
...
+# reload the modules
+rmmod kvm_intel kvm_amd kvm
+modprobe kvm_intel kvm_amd kvm
+
+KVM_SYSFS=/sys/module/kvm/parameters/
+KVM_OPTIONS="enable_vmware_backdoor force_emulation_prefix"
+
+for opt in $KVM_OPTIONS; do
+  if [ ! -f "$KVM_SYSFS/$opt" ]; then
+    echo "kernel option $opt not set" | tee -a $OUTPUTFILE
```
veruu: I only noticed this now, but should there be any result reporting before exiting here?

vm/kvm-unit-tests/runtest.sh
```diff
@@ -0,0 +1,118 @@
...
+
+# enable nmi_watchdog since some unit tests depend on this
+echo 0 > /proc/sys/kernel/nmi_watchdog
+
+# clone the upstream kvm-unit-tests
+git clone git://git.kernel.org/pub/scm/virt/kvm/kvm-unit-tests.git
+cd kvm-unit-tests
+git checkout 47cc3d85c2fe6fe4618e513b39046abe73f2803c
+if [ $? -ne 0 ]; then
+    exit 1
```
veruu: I believe this line should be put after the `rhts-abort` line, otherwise that code won't run

vm/kvm-unit-tests/runtest.sh
rasibley: @veruu yeah probably :-) I updated in both places

vm/kvm-unit-tests/runtest.sh
rasibley: @veruu your correct! good catch, it's probably not needed since we're setting it to warn/aborted, I just removed it.

veruu: But the script will continue to execute? It shouldn't change the Beaker result but would confuse people debugging the failures, maybe. At least I imagine

rasibley: @veruu since we're using rhts-abort, it will abort the recipe here, it shouldn't continue to run and the result will be reported as WARN: https://beaker-project.org/docs/user-guide/task-environment.html

So this is why I didn't think we needed 'exit 1' but I can add it back if you would like.

PR #44, opened 2018-11-27
blacklist/pass running usex on s390x

ftjahjad:
"usex" test will blacklist (PASSED), when running on s390x


standards/usex/1.9-29/runtest.sh
```diff
@@ -46,72 +46,77 @@ function VerboseCupsLog()
...
 
 # ---------- Start Test -------------
-RHELVER=""
-cat /etc/redhat-release | grep "^Fedora"
-if [ $? -ne 0 ]; then
-    RHELVER=$(cat /etc/redhat-release |sed 's/.*\(release [0-9]\).*/\1/')
-fi
+# verify to not run on s390x
+if [ "$(uname -i)" = "s390x" ]; then
+    rhts-report-result $TEST SKIP $OUTPUTFILE
```
veruu: Can we call `exit()` or something similar here, just so we don't need to change the rest of the code with the `else` clause?

ftjahjad:
I push the change to use "exit" instead of if-else.

Thanks,
-Fendy

PR #47, opened 2018-12-05
add metadata file to include Fedora dependencies

rasibley:
@veruu in order to enable usex on stable release we need to enable dependencies for Fedora, otherwise it will fail.


rasibley:
@veruu also I noticed this file was missing from master branch which is why I'm adding it, let me know if I need to do this another way.

veruu:
@rasibley we had to remove the metadata file to make SKIP work (for s390x). I don't remember the number of the jira ticket, but it should be in the board somewhere (or ask @ftjahjad ). The explanation of why we had to remove it is in the ticket.

Of course if you know about a solution that allows for the use of metadata file while having the skipping work please do let us know, but we weren't able to find one.

PR #58, opened 2019-01-10
add readme for xfstests

rasibley:


filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,28 @@
+xfstests test suite
+=====================================
+Wrapper for xfstests test suite, see http://git.kernel.org/?p=fs/xfs/xfstests-dev.git;a=summary \
+Test Maintainer: [Xiong Murphy Zhou](mailto:xzhou@redhat.com)
+
+How to run it
+-------------
+Make sure you have at least 25G free disk space in this partition
```
spbnick: Would be nice to mention why you'd need 25GB.

filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+$ df -H .
+```
+
+Clone the tests-beaker repository
+```bash
+$ git clone https://github.com/CKI-project/tests-beaker.git
```
spbnick: Since the README.md is already in the directory with the test, we don't need any instructions on how to clone the repo.

filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+Configure which test cases to run, e.g.
+```bash
+$ export TEST_PARAM_FSTYPE=xfs
+$ export TEST_PARAM_RUNTESTS=generic/001
+```
+
+Execute the test
```
spbnick: What's missing, I think, is the list of dependencies, where to get them, and how to install them. This is especially important for outside developers. The best way to produce such instructions would be to get a fresh install of e.g. Fedora and try to make the test work there.

If you'd rather target these instructions at internal developers only for now, then just a quick mention of what needs to be installed would be enough. Might be a good idea to ask an actual developer for a review :)

Lastly, we don't want to repeat all this (and probably some other things) for all the tests. So the common instructions could go into the README.md in the root of the repo, and the test README.md's could refer to the top one for pre-requisite instructions.

filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+$ cd filesystems/xfs/xfstests/
```
spbnick: Similarly, it's not a good idea to assume the global test location in a local README.md. Just drop this line :)

filesystems/xfs/xfstests/README.md
rasibley: I think @jencce could answer this question better

filesystems/xfs/xfstests/README.md
rasibley: Good point, I removed it

filesystems/xfs/xfstests/README.md
rasibley: Removed

filesystems/xfs/xfstests/README.md
rasibley: @jencce can you provide the details here regarding dependencies ? Also yes we can have a developer review the final version. I'm not sure if there will be any common instructions after these changes, other than make run, but we can move them to the root repo if so.

filesystems/xfs/xfstests/README.md
jencce: That's what we need for best coverage in beaker tests. It's not mandatory. We can remove it.

filesystems/xfs/xfstests/README.md
jencce: My thought is that, xfstests is a community project. We backport it to beaker from its upstream. Now we are trying to introduce it at another community project(if I understand it right -:),  we can just point to xfstests its own README:
https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git/tree/README

There is a dep-install.sh to install all dependencies.

All filesystem relative developers know xfstests, most of them are contributing to xfstests upstream.

spbnick: Sounds good, @jencce, then we can simply refer to the upstream instructions from here, if they're present in our repo. However, if we're going to instruct people to run the Beaker compatibility wrapper, then we gotta explain how to install the dependencies for it.

filesystems/xfs/xfstests/README.md
rasibley: I removed this

filesystems/xfs/xfstests/README.md
rasibley: I added more details based on @jencce response, @spbnick can you take another look ? I'm trying to leave beaker out of it completely and only provide simple basic steps anyone internal/external can run.

filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+### Configure which test cases to run, e.g.
+```bash
+$ export TEST_PARAM_FSTYPE=xfs
+$ export TEST_PARAM_RUNTESTS=generic/001
+```
+
+### Dependencies
+Dependencies will be automatically installed with dep-install.sh when you execute \
+'make run', however to see a complete list of dependencies, please see \
```
spbnick: Nitpick: I think it's gonna sound more natural if you write "however, for a complete list of dependencies, please see".

filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,22 @@
...
+https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git/tree/README
+
+
+### Execute the test
+```bash
+$ make run
```
spbnick: I tried this on a fresh Fedora 29 workstation install and I got this:
```
[nkondras@localhost xfstests]$ make run
Makefile:32: /usr/share/rhts/lib/rhts-make.include: No such file or directory
make: *** No rule to make target '/usr/share/rhts/lib/rhts-make.include'.  Stop.
```

This obviously needs Beaker stack prerequisites installed. If we're targeting internal users so far, then just mentioning that could be good enough. However, if we want anybody else from the outside to run the test like you specify here, then we need to make sure they can get everything required, and can *actually* run the tests.

If we want to target outside developers, then probably the most economical way (and closest to what we run) would be to direct them to run the Beaker wrapper for every test, and describe how to install the dependencies. An easy way to do this is to get a fresh Fedora install, download the tests, try to run, and document everything you had to do to make them work.

I actually don't see any other way to do that, since we're downloading upstream code and are patching it. So to get what we run they should do the same.

filesystems/xfs/xfstests/README.md
rasibley: I cleaned it up based on your suggestion.

filesystems/xfs/xfstests/README.md
rasibley: I completely forgot about restraint-rhts, I provisioned a system internally in beaker to run xfstests, and stripped it down to remove any restraint/beaker rpms and corresponding repos, then I did the following:
```
wget -O /etc/yum.repos.d/beaker-client.repo https://beaker-project.org/yum/beaker-client-Fedora.repo
wget -O /etc/yum.repos.d/beaker-harness.repo https://beaker-project.org/yum/beaker-harness-Fedora.repo
dnf install -y beaker-client beakerlib restraint-rhts
```
However it complained about conflicts and other failures, I must be missing other dependencies/repos. @jencce what other steps are we missing here?

spbnick: @rasibley, just install a Fedora 29 into a VM (e.g. in virt-manager) on your own host (it doesn't take long) and try to run tests there. This way you'll get cleanest instructions.

rasibley: Thanks @spbnick I did originally test on a vm using 1MinuteTip, but didn't think that was the right approach since I don't believe xfstests will run properly in a vm, maybe @jencce can confirm ?

rasibley: true :-) here is the complete log in a fresh VM install of F29 using virt-manager: https://paste.fedoraproject.org/paste/FpLt30acTogm6SubJVGrxw

I'll let @jencee confirm if we're missing anything.

jencce: xfstests can run in VM well and Racheal's log looks good enough.

rasibley: thanks @jencce I updated the root README with the common steps, @spbnick can you take a look ? :-)

README.md
```diff
@@ -1,6 +1,12 @@
 # Beaker tasks used with [skt](https://github.com/RH-FMK/skt) runner
 
-
+## How to run tests
+Here is a list of common prerequisites for all beaker tests, test specific dependencies and steps can be found in the README.md within the test directory.
```
spbnick: Could you put full stop in place of comma after "for all beaker tests"? Also put a dash in "test-specific", and write "within each test's directory"? I.e. write this:
> Here is a list of common prerequisites for all beaker tests. Test-specific dependencies and steps can be found in the README.md within each test's directory.

README.md
```diff
@@ -1,6 +1,12 @@
...
+~~~
+$ wget -O /etc/yum.repos.d/beaker-client.repo https://beaker-project.org/yum/beaker-client-Fedora.repo
+$ wget -O /etc/yum.repos.d/beaker-harness.repo https://beaker-project.org/yum/beaker-harness-Fedora.repo
+$ dnf install -y beaker-client beakerlib restraint-rhts
```
spbnick: You're using `$` as the shell prompt above. However, that's used by non-privileged users, for which all of the above commands would fail. You could either add `sudo` in front of those commands or switch the prompt to `#`. I'd prefer the former, as people don't normally use the root shell these days.

filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,21 @@
...
+## How to run it
+
+### Configure which test cases to run, e.g.
+```bash
+$ export TEST_PARAM_FSTYPE=xfs
+$ export TEST_PARAM_RUNTESTS=generic/001
+```
+
+### Dependencies
+Dependencies will be automatically installed with dep-install.sh when you execute \
```
spbnick: Could you add a sentence before this one saying to refer to the top-level README.md for common dependencies? People are going to land here from URLs in the reports and get confused, if they don't follow that and try to run the test.

filesystems/xfs/xfstests/README.md
```diff
@@ -0,0 +1,21 @@
+# xfstests test suite
+Wrapper for xfstests test suite, see http://git.kernel.org/?p=fs/xfs/xfstests-dev.git;a=summary \
+Test Maintainer: [Xiong Murphy Zhou](mailto:xzhou@redhat.com)
+
+## How to run it
+
+### Configure which test cases to run, e.g.
+```bash
+$ export TEST_PARAM_FSTYPE=xfs
+$ export TEST_PARAM_RUNTESTS=generic/001
```
spbnick: Could you put this section right before the "Execute the test" section? I think it relates to it closer than to the "Dependencies".

rasibley:
@spbnick thanks! could you please review again ?

PR #62, opened 2019-01-28
Minor update for kvm-unit-tests to support aarch64

idorax:
Fix for FASTMOVING-484, check test can be run on aarch64 or not.

### DIFF
--- a/vm/kvm-unit-tests/runtest.sh
+++ b/vm/kvm-unit-tests/runtest.sh
@@ -6,11 +6,45 @@ unset ARCH
 unset STANDALONE
 cpus=$(grep -c ^processor /proc/cpuinfo)
 
-# test can only run on hardware that supports virtualization 
-if egrep -q '(vmx|svm)' /proc/cpuinfo; then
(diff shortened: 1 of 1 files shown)


veruu:
Thanks Vector! Do you think you can maybe merge the arch checking that's below this check together with the function? Since the function is already checking for x86_64 and aarch64 specific flags, any other arches should be unsupported and we don't need that additional arch check, right?

idorax:
Thanks Veronica for your quick review and comments.  Let me remove L42-49 since the tests only support x86_64 and aarch64.

idorax:
Veronica, removing L42-49 is done, please take a look again, thanks!

veruu:
Thanks for a quick change! The test skipping message only mentions virt support on CPU and not the supported architectures, can they be added there so we don't lie in the logs please?

idorax:
Veronica, I reorganized the code, and also did a basic unit testing, the test log is saved in [1]. Please review again, thanks!
[1] http://pastebin.test.redhat.com/702936

veruu:
I haven't tested it myself but the changes look valid to me, thanks!

PR #64, opened 2019-01-30
Sanitize sanity test for loop device and move to the public repo

idorax:
This is to port test case `/kernel/loopdev/sanity` to public repo, and the latest iozone [1] is used. Please help to review, thanks!
[1] http://www.iozone.org/src/current/iozone3_487.tar

### DIFF
--- a/filesystems/loopdev/sanity/Makefile
+++ b/filesystems/loopdev/sanity/Makefile
@@ -1,0 +1,81 @@
+# The name of the package under test
+PACKAGE_NAME=kernel
+
+# The toplevel namespace within which the test lives.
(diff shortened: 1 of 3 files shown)


rasibley:
hi @veruu do you have time to review this one ? We'd like to include this in the upstream job :-) Once enabled here, we can test in kpet-db CI.

veruu:
I can't find any Beaker jobs that run the test to verify it really works, but don't see anything weird in the code itself. I'd merge it and we can always fix issues later.

rasibley:
thanks @veruu ! yes I couldn't either :( but this will be tested in kpet-db CI once I update to point to the upstream location, thanks for reviewing.

PR #65, opened 2019-01-30
Minor update for kvm-unit-tests to support aarch64 (cont)

idorax:
The harness (i.e. beaker) does clear dmesg, hence we have to fetch the output of kernel buffer from "journalctl -k" if the output of dmesg(1) is empty.


idorax:
A basic unit testing is done, please refer to http://pastebin.test.redhat.com/704557

rasibley:
This will fail on ARM when checking if the kernel options are set correctly, I can file a separate PR with these changes.

rasibley:
Now that https://github.com/CKI-project/tests-beaker/pull/67 was merged, this one can be closed.

PR #67, opened 2019-02-05
don't check the kernel options on ARM

rasibley:
This is a follow on to @idorax PR, additional changes needed for ARM testing


rasibley:
@cki-bot can you test this PR please?

rasibley:
hi @veruu can you review this one ? cki-bot found one related failure for kvm but it's fixed in a later version of qemu-kvm, the MR was merged to updated the pinned rhel8 compose to the latest, so we shouldn't see this error anymore. I'll forward you Luiz's email about it.

PR #68, opened 2019-02-06
fix lookaside for upstream

rasibley:
Minor update to fix lookaside variable to only point to upstream location


rasibley:
@veruu hi :-) There was one minor issue with this test, it tried to use the default lookaside var, I set it to only fetch from upstream, I'll send you a test job.

PR #75, opened 2019-02-13
kpkginstall fixes for RPM repos

veruu:


distribution/kpkginstall/runtest.sh
```diff
@@ -5,34 +5,25 @@
...
-# Status: zero if version is found, non-zero on failure
 function get_kpkg_ver()
 {
-  case "$KPKG" in
+  echo "Extracting kernel version from ${KPKG_URL}" | tee -a ${OUTPUTFILE}
+
+  case "${KPKG_URL}" in
     *.tar.gz)
-            tar tf "$1" | sed -ne '/^boot\/vmlinu[xz]-[1-9]/ {s/^[^-]*-//p;q}; $Q1'
+            KPKG=${KPKG_URL##*/}
```
spbnick: You need to add `declare` here to avoid creating a global variable, if you'd like that.
I.e.: `declare KPKG=${KPKG_URL##*/}`, or even better `declare -r KPKG=${KPKG_URL##*/}` for a read-only variable. Also, lowercase, since it's not global: `declare -r kpkg=${KPKG_URL##*/}`.

And then you'd probably want to remove passing the argument to this function in the code below, since you're using global state now (KPKG_URL and the name of the dnf repo).

distribution/kpkginstall/runtest.sh
```diff
@@ -5,34 +5,25 @@
...
-            karray=(${1//-/ })
-            if [ "${#karray[@]}" == "3" ]; then
-              echo "${karray[1]}-${karray[2]}.$ARCH"
-            elif [ "${#karray[@]}" == "4" ]; then
-              # uname -r have variants at the end
-              KVAR="${karray[1]}"
-              echo "${karray[2]}-${karray[3]}$KVAR.$ARCH"
-            fi
+            # Grab the kernel version from the provided repo directly
+            dnf --disablerepo="*" --enablerepo="kernel-cki" list kernel | grep kernel-cki | awk '{print $2}'
```
spbnick: You could also write `awk '/kernel-cki/ {print $2}'` instead of `grep kernel-cki | awk '{print $2}'`.

distribution/kpkginstall/runtest.sh
```diff
@@ -5,34 +5,25 @@
...
             ;;
   esac
 }
 
 function targz_install()
 {
+  KPKG=${KPKG_URL##*/}
```
spbnick: Same thing here with local variables.

veruu:
@spbnick all valid points, I knew you're the right person for this review :) I made the changes, please check if I didn't miss anything

spbnick:
This looks better :) There are still other unnecessary-global variables, but we can fix them later.

spbnick:
Please go ahead and merge, if you've tested it :)

veruu:
Yeah, I went with the path of the least resistance in order to make the existing code working with the RPM repos. Maybe we'll find out some other things are missing but the parts I tested on my machine worked fine.

PR #98, opened 2019-02-26
initial commit of mod_ssl smoke test

ftjahjad:
@veruu  and @rasibley 

Please review mod_ssl  test initial commit PR

### DIFF
--- a/packages/httpd/mod_ssl-smoke/runtest.sh
+++ b/packages/httpd/mod_ssl-smoke/runtest.sh
@@ -50,8 +50,8 @@ rlJournalStart
 	    if [ -e "/etc/system-fips" ] && grep -q 1 /proc/sys/crypto/fips_enabled; then
                 rlPass "fips mode enabled"
 	    else
-		rlFail "fips mode disabled. Test requires fips mode!"
-		rlDie "Skipping the test!"
+                echo "fips mode disabled. Test requires fips mode! Skipping." | tee -a $OUTPUTFILE
+                rhts-report-result $TEST SKIP $OUTPUTFILE
 	    fi
         fi
         rlAssertRpm --all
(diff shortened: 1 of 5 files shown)


packages/httpd/mod_ssl-smoke/runtest.sh
```diff
@@ -0,0 +1,100 @@
...
+fi
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        if [[ $FIPS == "true" ]]; then
+	    if [ -e "/etc/system-fips" ] && grep -q 1 /proc/sys/crypto/fips_enabled; then
+                rlPass "fips mode enabled"
+	    else
+		rlFail "fips mode disabled. Test requires fips mode!"
```
veruu: Hi, should this be skipping the test instead of failing? Asking because the line below specifically says "Skipping the test" and the command doesn't match that.

ftjahjad: @veruu 
Thanks for the review  and totally agree with your comment.
I update the code , please re-review again.

veruu: Thanks for the update! Just a question since I'm not familiar with this type of test - does the `rhts-report-result` assure the rest of `rhPhase`s don't run? I'm just confused about mixing those two together and what effects it has

ftjahjad: Thanks for the review.

I  trigger to make "skipping"  by adding <params "FIPS" values="true" ...>  .
 rhts-report-result  reports "Skipped" and the recipe existed without issue/effect.
this  is "skipped" beaker job:# 3369892
and 
"passed" beaker job:# 3369899
 
*added missing "exit" after  rhts-report-result .

veruu: Thanks for the changes! I'll squash the commits into one.

PR #100, opened 2019-03-07
reduce the number of ltp-lite case in CKI

wangli5665:
As the reason for CKI is to provide a quick gating test for each
pre-apply patch, so better to run some stable and functional case
in this ltp-lite. Here I just remove many stress and fragile cases
from the runtest file to make ltp-lite has a pretty baseline.

Btw, we will also run the complete ltp-lite in kernel Tier1 and other
deeping test. So we do have lost nothing but just making CKI more faster
and stable.

Signed-off-by: Li Wang <liwang@redhat.com>


veruu:
After discussions with @rasibley , merging!

PR #104, opened 2019-03-11
inital commit of audit-testsuite

ftjahjad:
@veruu 

Please review  audit-testsuite PR request.

### DIFF
--- a/packages/audit/audit-testsuite/runtest.sh
+++ b/packages/audit/audit-testsuite/runtest.sh
@@ -1,0 +1,157 @@
+#!/bin/bash
+# vim: dict=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   runtest.sh of /CoreOS/audit/Sanity/audit-testsuite
+#   Description: Execute audit-testsuite
+#   Author: Ondrej Moris <omoris@redhat.com>
+#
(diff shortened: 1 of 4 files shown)


packages/audit/audit-testsuite/runtest.sh
```diff
@@ -0,0 +1,157 @@
...
+
+        rlRun "pushd audit-testsuite" 0 || rlDie
+        rlRun "git checkout $GIT_BRANCH" 0 || rlDie
+
+        # Apply workarounds for beaker environment.
+        rlRun "unset DISTRO" 0
+        rlRun "echo $(id -u) >/proc/self/loginuid" 0
+
+	# Turn off intel-only test on non-intel architectures.
+	test "$(rlGetPrimaryArch)" != "x86_64" && \
```
veruu: I'm a bit confused by this check -- the comment says "intel-only" but x86_64 covers AMD too? Either the comment should be updated or the code is wrong and the test will fail on AMD machines.

Other than this I don't see anything weird in the test, but I'm not familiar with it :)

ftjahjad: That comment is from  original internal repo source code., just realize the comment is misleading  :-) 

This the full explanation:
When  running on  Non arch:x86_64. the specific test must be removed since only supported on x86_64  architecture only. :-D

I can change  the comment to "# Turn off x86_64 specific test  when running on non x86_64 architectures"

is that ok?

veruu: Yes that works :) thanks for explanation!

ftjahjad: @veruu, Thanks :-D

Please re-review again the updated comment.

PR #106, opened 2019-03-12
kdump: Added 3 kdump tests

emmawu:
Three kdump tests:
1) setup-bare-metal
    test if crashkernel is reserved and kdump service is started.
2) crash-sysrq-c
    trigger sysrq-c panic and test kdump procoss
3) analyse-basic
    check if a vmcore is saved from kdump

### DIFF
--- a/kdump/analyse-basic/runtest.sh
+++ b/kdump/analyse-basic/runtest.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2019 Red Hat, Inc.
+#
+#   This program is free software: you can redistribute it and/or
+#   modify it under the terms of the GNU General Public License as
+#   published by the Free Software Foundation, either version 2 of
+#   the License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be
+#   useful, but WITHOUT ANY WARRANTY; without even the implied
+#   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#   PURPOSE.  See the GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program. If not, see http://www.gnu.org/licenses/.
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+# Source kdump common functions
+. ../include/runtest.sh
+
+analyse()
+{
+    GetCorePath
+    LsCore
+}
+
+#+---------------------------+
+
+Multihost "analyse"
--- a/kdump/crash-sysrq-c/runtest.sh
+++ b/kdump/crash-sysrq-c/runtest.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2019 Red Hat, Inc.
+#
+#   This program is free software: you can redistribute it and/or
+#   modify it under the terms of the GNU General Public License as
+#   published by the Free Software Foundation, either version 2 of
+#   the License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be
+#   useful, but WITHOUT ANY WARRANTY; without even the implied
+#   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#   PURPOSE.  See the GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program. If not, see http://www.gnu.org/licenses/.
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+# Source kdump common functions
+. ../include/runtest.sh
+
+function Crash() {
+    DisableAVCCheck
+
+    if [ ! -f "${K_REBOOT}" ]; then
+        PrepareReboot
+
+        local sysrq_value
+        sysrq_value=$(cat /proc/sys/kernel/sysrq)
+        [ "$sysrq_value" -eq 0 ] && Warn "kernel.sysrq is set to 0 which is unexpected."
+
+        # Crash the system.
+        touch "${K_REBOOT}"
+        Report 'boot-2nd-kernel'
+        sync
+
+        # Check kdump status before triggering panic
+        CheckKdumpStatus
+
+        # Trigger panic
+        echo c >/proc/sysrq-trigger
+        # Should stop here.
+    else
+        rm -f "${K_REBOOT}"
+    fi
+}
+
+# --- start ---
+Multihost "Crash"
--- a/kdump/include/runtest.sh
+++ b/kdump/include/runtest.sh
@@ -1,0 +1,1015 @@
+#!/bin/bash
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2019 Red Hat, Inc.
+#
+#   This program is free software: you can redistribute it and/or
+#   modify it under the terms of the GNU General Public License as
+#   published by the Free Software Foundation, either version 2 of
+#   the License, or (at your option) any later version.
+#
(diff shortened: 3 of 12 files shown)


emmawu:
Trigger pattern:

./kernel/kexec*
./include/linux/kexec*
./arch/x86/*kexec*
./arch/x86/kernel/crash.*
./arch/powerpc/*kexec*
./arch/powerpc/kernel/crash.*
./arch/s390/*kexec*
./arch/arm64/*kexec*
./arch/arm64/kernel/crash_dump.c
./arch/s390/kernel/crash_dump.c
./kernel/crash_core.c
./fs/proc/vmcore.c

Please note these 3 tests have to be run in exact order:
- setup-bare-metal
- crash-sysrq-c
- analyse-basic

And '\<watchdog panic="ignore"/>' has to be set to the beaker job as triggering a system panic is part of the test.

Thanks,
Emma

kdump/include/runtest.sh
```diff
@@ -0,0 +1,1019 @@
...
+{
+    local core=$1
+    [ -z "$core" ] && return 1
+
+    if [[ "${K_NFSSERVER}" =~ ^[[:space:]]*$ ]]; then
+        # NFSSERVER=ibm-x3250m4-03.rhts.eng.pek2.redhat.com
+        # VMCOREPATH=/data/vmcore
+        # This Server is maintained by cye@redhat.com located in PEK2 Lab
+
+        K_NFSSERVER="ibm-x3250m4-03.rhts.eng.pek2.redhat.com"
```
rasibley: hi @emmawu can we instead pass this via beaker task param ? this way we don't expose the internal hostname publically ?

spbnick:
If these *must* be executed in sequence and don't make sense without each other, can we make them a single task and invoke them from a script instead? From the POV of kpet, each task is a single independent test, and supporting multi-task tests would be an unnecessary burden to kpet code.

kdump/include/runtest.sh
emmawu: Thanks! I've corrected it. It would return is no K_NFSSERVER is provided.

emmawu:
> If these _must_ be executed in sequence and don't make sense without each other, can we make them a single task and invoke them from a script instead? From the POV of kpet, each task is a single independent test, and supporting multi-task tests would be an unnecessary burden to kpet code.

Hi Nikolai. 
Kdump panic test may reboot system up to 2 times (or one time depends on test machine configurations). We tried to handle one reboot in one single test. Otherwise it has to check whether system is currently rebooted once or twice or not rebooted yet, and add/remove flags accordingly. The test process control would be more complicated. Can we keep it in 2 separate cases? That is setup-bare-metal and crash-sysrq-c? I could merge the analyse-basic to crash-sysrq-c.

spbnick:
I see. Yes, reboot handling can be hairy. Still, I would prefer it in a single task, and I could help you merge those three. Otherwise the complexity will eventually creep into all parts of our pipeline, not just kpet. Could you point me to the parts of the test code that are troublesome?

emmawu:
And actually, putting analyse-basic is only the first step of moving kdump tests to KPET. The idea is moving a more though test analyse-vmcore() to read the dumped vmcore. In this case, no panic/reboot is expected. And it would report failure if any unexpxected panic/reboot detected.
So basically, 
setup-bare-metal - may reboot once depends on system configuration
crash-sysrq-c - expected system panic and reboot once.
analyse-vmcore - no reboot/panic is expected. report failure if any reboot is detected.

kdump/setup-bare-metal/runtest.sh
```diff
@@ -0,0 +1,196 @@
...
+        fi
+    fi
+
+    echo "$args"
+}
+
+
+SetupKdump()
+{
+    if [ ! -f "${K_REBOOT}" ]; then
```
emmawu: This is the place where a system reboot is detected.

kdump/setup-bare-metal/runtest.sh
```diff
@@ -0,0 +1,196 @@
...
+        # Ensure Kdump Kernel memory reservation
+        grep -q 'crashkernel' <<< "${KER1ARGS}" || {
+            local kdumpMem=$(DefKdumpMem)
+            [ -z "${KER1ARGS}" ] || kdumpMem=" ${kdumpMem}"
+
+            $IS_RHEL5 && KER1ARGS+="${kdumpMem}" || {
+                # memory >= auto-threshold and kdump already default on
+                [ `cat /sys/kernel/kexec_crash_size` -eq 0 ] && {
+                    Log "`grep MemTotal /proc/meminfo`"
+                    KER1ARGS+="${kdumpMem}"
```
emmawu: if no kexec_crash_size is reserved after boot,  it will set a value to KER1ARGS, which will be later pass to  grubby --args to update kernel options.

kdump/setup-bare-metal/runtest.sh
```diff
@@ -0,0 +1,196 @@
...
+            Log "- Changing boot loader."
+            {
+                /sbin/grubby                     \
+                    --args="${KER1ARGS}"         \
+                    --update-kernel="${default}" &&
+                if [ "${K_ARCH}" = "s390x" ]; then zipl; fi
+            } || FatalError "Error changing boot loader."
+
+            Report 'pre-reboot'
+            Log "- Rebooting\n"; sync; rhts-reboot
```
emmawu: if KER1ARGS is set in previous steps it would call grubby to update kernel option and manage a reboot.

kdump/crash-sysrq-c/runtest.sh
```diff
@@ -0,0 +1,52 @@
...
+        touch "${K_REBOOT}"
+        Report 'boot-2nd-kernel'
+        sync
+
+        # Check kdump status before triggering panic
+        CheckKdumpStatus
+
+        # Trigger panic
+        echo c >/proc/sysrq-trigger
+        # Should stop here.
```
emmawu: This echo c would cause a system panic. A ${K_REBOOT} has been created to mark the system will be rebooted.

kdump/crash-sysrq-c/runtest.sh
```diff
@@ -0,0 +1,52 @@
...
+    else
+        rm -f "${K_REBOOT}"
```
emmawu: Remove the  reboot flag after system reboot.

kdump/analyse-basic/runtest.sh
```diff
@@ -0,0 +1,33 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+# Source kdump common functions
+. ../include/runtest.sh
+
+analyse()
+{
+    GetCorePath
+    LsCore
```
emmawu: This is a basic test for now. But later if it's changed to analyse-vmcore.
We have another function to flag and check system status.  like thits.
CheckUnexpectedReboot()
{
    if [ -f "$K_REBOOT" ]; then
        rm -f "$K_REBOOT"
        MajorError "- Unexpected reboot is detected. Please check if system has \
been rebooted from panic or other possible incidents."
    fi
    touch "$K_REBOOT"
    sync;sync;
}

emmawu:
Hi Nikolai,
I don't know how i could point your where the current process control is...Can you see the comment I added in above review sessions ?

rasibley:
hi @spbnick the kdump test(s) will run separately using your recent multi-host feature in kpet, what burden would it be to kpet code, we can include multiple tasks and we already do this with xfstests.

spbnick:
We have one task element per test case so far (in xfstests too), and if possible I'd prefer to keep it that way. This will allow us to generate those task elements automatically in the future, so we can move the list of maintainers, waived status and other task data from XML to YAML, where it would be more manageable.

We can try to pretend that those three are separate cases, but we would need to keep their parameters in sync manually, and then that's gonna break when we introduce test order randomization.

rasibley:
I can see randomization breaking things, would it be difficult to exclude tests which are run stand alone on a separate host like kdump tests ? Is it possible to merge this now (pending testing) until we find a better way to combine these tasks into a single case ?

spbnick:
If we can, it would be good to avoid merging this as separate tasks.
We can't use it yet anyway, because we don't have support for RPM-only tests in kpet yet.
However, if by the time we do we can't make this into a single task, it will work as is for a while.

emmawu:
Hi, @spbnick @rasibley 
I re-worked the test structure.  Now it's in one test case. I filed another PR: https://github.com/CKI-project/tests-beaker/pull/110

spbnick:
Thank you very much, Emma! Sorry I couldn't come and help soon enough. I'm glad that you managed to put this into a single task. However, if you, or @rasibley know of any tests that really cannot be put into a single task (e.g. some extensive multihost tests which require multiple tasks per host) please let me know, so that we prepare to handle that in kpet after all and not ask anymore extra work from people. Thank you.

spbnick:
Superseded by #110, closing.

spbnick:
@emmawu, RPM-only tests are e.g. this kdump test which requires some information specific to RPM-packaged kernels, which are only available in our Brew pipeline.

PR #107, opened 2019-03-12
kpkg fixups for fedora

veruu:


distribution/kpkginstall/runtest.sh
```diff
@@ -153,7 +153,7 @@ EOF
   fi
 
   # The package was renamed (and temporarily aliased) in Fedora/RHEL
-  if $YUM search kernel-firmware | grep kernel-firmware ; then
+  if $YUM search kernel-firmware | grep "^kernel-firmware\.noarch" ; then
```
spbnick: This will work fine, but just in general, to keep it simpler I'd use single quotes here. Compare:

```
       Enclosing  characters  in  single quotes preserves the literal value of
       each character within the quotes.  A single quote may not occur between
       single quotes, even when preceded by a backslash.
```
vs:
```
       Enclosing  characters  in  double quotes preserves the literal value of
       all characters within the quotes, with the exception of $, `,  \,  and,
       when history expansion is enabled, !.  When the shell is in posix mode,
       the ! has no special meaning within double quotes,  even  when  history
       expansion  is  enabled.   The  characters  $ and ` retain their special
       meaning within double quotes.  The backslash retains its special  mean‐
       ing  only when followed by one of the following characters: $, `, ", \,
       or <newline>.  A double quote may be quoted  within  double  quotes  by
       preceding  it  with a backslash.  If enabled, history expansion will be
       performed unless an !  appearing in double quotes is  escaped  using  a
       backslash.  The backslash preceding the !  is not removed.
```

PR #110, opened 2019-03-14
kdump: Add a kdump-sysrq-c test

emmawu:
It tests
1) crashkernel memory reservation
2) kdump service startup
3) Vmcore saving after a system panic

### DIFF
--- a/kdump/include/runtest.sh
+++ b/kdump/include/runtest.sh
@@ -1,0 +1,595 @@
+#!/bin/bash
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2019 Red Hat, Inc.
+#
+#   This program is free software: you can redistribute it and/or
+#   modify it under the terms of the GNU General Public License as
+#   published by the Free Software Foundation, either version 2 of
+#   the License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be
+#   useful, but WITHOUT ANY WARRANTY; without even the implied
+#   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#   PURPOSE.  See the GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program. If not, see http://www.gnu.org/licenses/.
+#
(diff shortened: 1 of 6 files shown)


emmawu:
Trigger pattern:

./kernel/kexec*
./include/linux/kexec*
./arch/x86/kexec
./arch/x86/kernel/crash.*
./arch/powerpc/kexec
./arch/powerpc/kernel/crash.*
./arch/s390/kexec
./arch/arm64/kexec
./arch/arm64/kernel/crash_dump.c
./arch/s390/kernel/crash_dump.c
./kernel/crash_core.c
./fs/proc/vmcore.c

kdump/kdump-sysrq-c is the test case.
kdump/include is the common func lib

rasibley:
Thank you @emmawu ! I just tested it in beaker and it works well, @spbnick and @veruu any further comments, can we merge this ?

spbnick:
I didn't look into the code, but I'm happy Emma got it to just one task :)

rasibley:
Thanks @spbnick, @veruu are we ok to merge this one ?

kdump/include/runtest.sh
```diff
@@ -0,0 +1,595 @@
...
+
+    elif $IS_RHEL8; then
+        if   [[ "${K_ARCH}"  = "x86_64" ]]; then args="crashkernel=160M"
+        elif [[ "${K_ARCH}"  = "s390x"  ]]; then args="crashkernel=160M"
+        elif [[ "${K_ARCH}"  = ppc64*  ]]; then
+            args="crashkernel=0M-4G:384M,4G-16G:512M,16G-64G:1G,64G-128G:2G,128G-:4G"
+        elif [[ "${K_ARCH}"  = "aarch64"  ]]; then args="crashkernel=512M"
+        fi
+
+    elif $IS_RHEL5; then
```
veruu: Checking all the `if`s here, are we planning to run this task on Fedora kernels?

spbnick:
I haven't looked at the code of the test, and I'll leave it up to you, @rasibley and @veruu :)
Also, kpet should now be ready for this test (can run on a separate host with panic detection disabled and for pipelines with kernels in a repo only, i.e. Brew pipeline). Please take over the corresponding kpet-db MR after this is merged, so we can finally get this running.

kdump/include/runtest.sh
rasibley: @veruu I think it's ok if we do run on the Fedora kernel since the base compose is rhel8, could we go ahead and merge and make changes later if needed ?

veruu: Sure, that's fair

PR #117, opened 2019-03-19
initial commit of kern-mod-symb-req

rasibley:
@veruu can you please review ?


packages/redhat-rpm-config/kernel-module-symbol-requires/runtest.sh
```diff
@@ -0,0 +1,83 @@
...
+        rlRun "mv *.spec ~/rpmbuild/SPECS"
+    rlPhaseEnd
+
+    rlPhaseStartTest
+        # for proper rebuild of non-x86_64 arches ARCH has to be unset
+        ARCHBACKUP=$ARCH
+        unset ARCH
+        uname -a
+        rpm -q kernel kernel-devel
+        KVER=$(rpm -q --qf '%{VERSION}' kernel-devel)
```
veruu: Will this always work? I mean, we can't guarantee that there will be only one kernel-devel package installed on the system, can we? eg. it will be installed for the base kernel in kickstart

rasibley: @veruu it seems to pass everywhere in the brew pipeline, it grabs the latest kernel/kernel-devel. I"ll include example jobs in the ticket.

rasibley: Yes I believe it's a custom dependency for the test, I can double check with @ksrot

veruu: But the only reason it passes is because there's only one kernel-devel package, no? What happens if we eg. onboard Stanislav's test which grabs another kernel from brew to determine kabi changes and that one needs to install the devel package of that kernel, or the default kickstart changes and installs kernel-devel of the GA?

rasibley:
@veruu actually no there's two kernel-devel rpms installed, you can see in the logs I attached to the ticket when the test runs:
```
$ rpm -q kernel kernel-devel
kernel-3.10.0-957.el7.x86_64
kernel-3.10.0-1028.el7.x86_64
kernel-devel-3.10.0-957.el7.x86_64
kernel-devel-3.10.0-1028.el7.x86_64
```
and here we see it's using the sources from the latest kernel-devel:

```
make -C /lib/modules/3.10.0-1028.el7.x86_64/build M=/root/rpmbuild/BUILD modules
make[2]: Entering directory `/usr/src/kernels/3.10.0-1028.el7.x86_64'
  CC [M]  /root/rpmbuild/BUILD/testprov.o
  Building modules, stage 2.
  MODPOST 1 modules
  CC      /root/rpmbuild/BUILD/testprov.mod.o
  LD [M]  /root/rpmbuild/BUILD/testprov.ko
make[2]: Leaving directory `/usr/src/kernels/3.10.0-1028.el7.x86_64'
make[1]: Leaving directory `/root/rpmbuild/BUILD'
+ exit 0
```
not sure about the new tests onboarded by Stanislav, as long as the is considering the latest version for the kernel and kernel dependencies, but I may be missing something.

veruu:
@rasibley thanks for the example! Not sure about Stanislav's test either, I was just giving an example where it may be a problem. Because when we install older zstream kernels on ystream distro, the kernel versions we care about are older than the GA ones. Thus, based on what you say, we'll grab the version of the GA devel package if it was installed, not the one we want

rasibley:
I see, well looking at the spec file it sets the version to running version of the kernel `%define kver %(uname -r)`, I'd be happy to test this out on a z-stream job to see if it blows up :-)

veruu:
> I'd be happy to test this out on a z-stream job to see if it blows up :-)

Sounds fair!

In any case, the query seems to be working differently than on my machine (running `rpm -q --qf '%{VERSION}' kernel` returns `4.20.114.20.144.20.15` when I'm running on 4.20.14-200.fc29.x86_64 because I have multiple kernels installed)

packages/redhat-rpm-config/kernel-module-symbol-requires/runtest.sh
kkaarreell: Hi,
we do not really have experience with building kernel modules and therefore we are trying to use the simplest method available. The only purpose is to ensure that we are building against kernel version which has kernel-devel installed. You are right that with more packages installed it won't work properly, obviously this was not the case we would be hitting in our testing (as we do not touch kernel). Fell free to change this line to whatever you find appropriate, I would add | head -n 1 to our test source. Thanks for pointing this out and reaching out to me.

veruu: thanks for the comment! `uname -r` sounds good to me

rasibley:
Thanks for your feedback @kkaarreell,  @veruu I updated it to set the running kernel using 'uname -r'

veruu:
Thanks @rasibley ! I'll merge it now and we can do additional testing in kpet

PR #125, opened 2019-03-25
Add RT related cases to CKI testing.

linuxqiao:
Follow our documents, all the test passed by beaker [1].
And for "suggestion pattern", because the rt kernel is another flavour for rhel kernel. So we have no special "pattern" for rt kernel.

[1] https://beaker.engineering.redhat.com/jobs/3424443

### DIFF
--- a/rt-tests/include/Makefile
+++ b/rt-tests/include/Makefile
@@ -1,0 +1,50 @@
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2019 Red Hat, Inc.
+#
+#   This program is free software: you can redistribute it and/or
+#   modify it under the terms of the GNU General Public License as
+#   published by the Free Software Foundation, either version 2 of
+#   the License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be
+#   useful, but WITHOUT ANY WARRANTY; without even the implied
+#   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#   PURPOSE.  See the GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program. If not, see http://www.gnu.org/licenses/.
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+export TEST        := rt-tests/include
+export TESTVERSION := 1.0
+
+.PHONY: build run clean
+
+FILES := $(METADATA) runtest.sh Makefile
+
+build:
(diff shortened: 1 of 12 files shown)


linuxqiao:
@rasibley This is Qiao Zhao, and work on RT sst. Could you help to review this PR and let me know if I miss something?

rasibley:
hi @linuxqiao what type of hardware can we run these tests on, are we limited or can we run them anywhere ? Does the BIOS need to be tuned on the system to enable RT  ?

linuxqiao:
Hi @rasibley for rt kernel testing, no hardware request, we can run them anywhere(x86_64). 
For BIOS enable: no, we only need to run on a normal system for CKI testing with kernel-rt.

rasibley:
Thanks @linuxqiao I think we may need some additional changes in kpet-db to be able to add this test to RT testing, right now we trigger off the baseline RT tree and run all existing tests we have in CKI, however since this test is limited to RT kernel, we would need a way to include this test to the beaker xml for RT testing, but exclude it everywhere else. @spbnick can you comment ?

rasibley:
@linuxqiao I filed a ticket to track adding this feature, but for now I think we can still merge it, @veruu did you have any other feedback or can we merge this while we wait on this kpet feature ?

linuxqiao:
@rasibley thank you! please reach out to me if you need other.

rasibley:
@linuxqiao just thinking out loud, we could onboard this test right now, we could add a skip option to the test if not running with a rt kernel ? I see the test actually passes right now in the regular pipeline but that's because it installs the rt kernel which relates to the base compose under test, so it would be kind of meaningless since we're not testing with an updated rt-kernel @veruu do you think that's a good or bad idea :-)

rasibley:
Great! @linuxqiao can I ask you favor to please update your PR to *not* install rt-kernel since it's not needed since we already install the updated rt-kernel, then before you do do anything skip the test if not running on rt kernel, similar to this: https://github.com/CKI-project/tests-beaker/blob/master/vm/kvm-unit-tests/runtest.sh#L57

linuxqiao:
@rasibley I have updated the test case to match your suggestion:
1) remove "kernel-rt" package installation.
2) skip all rt related test on non-rt kernel.
Please review and let me know if need other updates. @veruu

rasibley:
@linuxqiao looks good ! Does env_test *need* to be run before the other tests to ensure the pkgs are installed ? Right now we're working on randomizing our tests so that would be an issue if so. If the tests need to be run in a certain order can you either add the dependencies to the other tests so they can be run stand alone or add a wrapper script to combine the tests ?

linuxqiao:
@rasibley you are right, "env_test" need to be run before the other tests. Now, I removed it and wrote a library "include/". Now all test case can be run stand alone. 
Please review and let me know if have any problem.

rasibley:
Thanks @linuxqiao no further comments, I tested it on both rhel7 and rhel8 including the skip option, details are in the ticket, @veruu can you review/merge ? :-)

PR #126, opened 2019-03-25
perf: adding internal-testsuite wrapper for CKI

frozenstein:
This is a wrapper for perf-test which whitelists some known failures
and unstable tests for the sake of having it stable and passing.


rasibley:
Thank you @frozenstein ! Are there any trees/arches we should exclude this testing or can we test everywhere ? Do you have a list of kernel patterns/files we should trigger testing against when testing kernel patches ?

frozenstein:
https://github.com/frozenstein/tests-beaker/commit/7effbfb2cb4c3158c4d7cee046b90114e4288f04

Fixes the ppc64le fails.

PR #130, opened 2019-03-27
Enable kvm self tests

idorax:
There is a case 'dirty_log_test' failing to be built, I just comment it out for the time being.

### DIFF
--- a/vm/kvm-self-tests/Makefile
+++ b/vm/kvm-self-tests/Makefile
@@ -1,0 +1,63 @@
+#
+# Copyright (c) 2019 Red Hat, Inc. All rights reserved.
+#
+# This copyrighted material is made available to anyone wishing
+# to use, modify, copy, or redistribute it subject to the terms
+# and conditions of the GNU General Public License version 2.
+#
+# This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied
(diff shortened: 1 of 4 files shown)


rasibley:
Thank @idorax it runs well :-) can you please update the test to skip where it's not supported like this ? https://github.com/CKI-project/tests-beaker/blob/master/vm/kvm-unit-tests/runtest.sh#L54

idorax:
Hi @rasibley, I have updated the related code according to your suggestion, please take a look again, thanks!

rasibley:
Thanks @idorax can you please exit after skip without 'return 1' otherwise the test will skip/abort, we want result="Skip" status="Completed" for all scenarios, similar to https://github.com/CKI-project/tests-beaker/blob/master/vm/kvm-unit-tests/runtest.sh#L55

idorax:
Hi @rasibley, the related code is updated, I use `exit 0` in rlSkip() explicitly.

idorax:
Hi @rasibley, I updated the test code by patching `tools/include/linux/bitmap.h` so as to enable case 'dirty_log_test'. If it is fixed in the future, we'll remove the patch accordingly. Please review again, thanks!

rasibley:
Thanks @idorax looks good now! I see it passes/skipped correctly now, I'll leave it to @veruu to review/merge.

PR #131, opened 2019-03-28
update stress-ng test:  pass when returns exit: 0,2,3  and exclude  two tests for rhel7 ppc64 

ftjahjad:
@rasibley , and @veruu 

Discuss with Test-maintainer and agree  that stress-ng is "pass" when returns with exit: 0,2,3, 
and  adding  two tests:fnctl/kcmp to   rhel7 ppc64  RHEL specific excludes.


rasibley:
Thanks @ftjahjad looks good to me, @veruu I believe @ftjahjad added example jobs in the jira ticket if your ok with merging

PR #132, opened 2019-03-29
kpkginstall fixups

veruu:


distribution/kpkginstall/runtest.sh
```diff
@@ -169,17 +202,17 @@ function rpm_install()
     echo "Kernel version is ${KVER}" | tee -a ${OUTPUTFILE}
   fi
 
-  $YUM install -y "kernel-$KVER" >>${OUTPUTFILE} 2>&1
+  $YUM install -y "$PACKAGE_NAME-$KVER" >>${OUTPUTFILE} 2>&1
```
major: You wrapped curly braces around `$PACKAGE_NAME` on the other lines. Should we do it here as well to keep it consistent?

PR #137, opened 2019-04-03
stress/stress-ng: abort if a reboot is detected

jbastian:
If a stressor causes a kernel panic and reboot, then abort the test
run.  Do not try to start the test again or it will likely get into
a panic-and-reboot-loop until the external watchdog kills it.


ftjahjad:
@jbastian , Thanks

@cki_bot , pleas test this PR

PR #140, opened 2019-04-04
Add dummy test

Toaster192:
Adding just a really stupid dummy test as a template


Toaster192:
@veruu Would this be alright?

PR #141, opened 2019-04-04
header-test: fix ppc64/ppc64le symlinks

jbastian:
This test fails with a Python error "NotADirectoryError: [Errno 20] Not a directory: 'el8/ppc64'", so swap the directory and symlink around to make Python happy.


rasibley:
Thanks @jbastian ! @veruu can we merge this one ?

PR #148, opened 2019-04-10
disable loop/001 for upstream 5.0.0

yizhanglinux:
Signed-off-by: Yi Zhang <yizhan@redhat.com>


rasibley:
hi @veruu can we merge this ?

PR #150, opened 2019-04-10
stress-ng: make efivar a known issue for all arches

jbastian:
efivar is triggering panics on some x86_64 systems too, so blacklist it for all architectures.


ftjahjad:
@jbastian 
 Thanks

@veruu 
please review this PR request

PR #161, opened 2019-04-16
kpkginstall: More verbose, use tee consistently

major:

### DIFF
--- a/distribution/kpkginstall/runtest.sh
+++ b/distribution/kpkginstall/runtest.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -x
+
 
 . /usr/bin/rhts_environment.sh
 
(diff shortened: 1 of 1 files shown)


distribution/kpkginstall/runtest.sh
```diff
@@ -62,7 +64,7 @@ function targz_install()
 {
   declare -r kpkg=${KPKG_URL##*/}
   echo "Fetching kpkg from ${KPKG_URL}" | tee -a ${OUTPUTFILE}
-  curl -OL "${KPKG_URL}" >>${OUTPUTFILE}
+  curl -OL "${KPKG_URL}" 2>&1 | tee -a ${OUTPUTFILE}
```
agustinhenze: if curl throws an error, the output file will be corrupted and we won't be able to see the error as well

agustinhenze: ok, we will see the error because you are using tee. But the output file will be corrupted as happened in the past

major: @agustinhenze Should we avoid writing stderr to outputfile at all? I was still trying to figure out what it was being used for.

agustinhenze: OK, I was wrong. OUTPUTFILE is what beaker-restraint shows as output. Probably would be better doing this change here https://github.com/CKI-project/tests-beaker/blob/master/distribution/kpkginstall/Makefile#L28 and not over each command, but it's already merged :man_shrugging: hehe

These are the url that I found trying to understand
https://beaker-project.org/docs/user-guide/task-environment.html
https://github.com/beakerlib/beakerlib/wiki/man

PR #162, opened 2019-04-17
stress-ng: switch to a whitelist of stressors

jbastian:
The class+blacklist method was growing too complicated, and it was
difficult to debug, so switch to a whitelist of stressors and log the
result of each stressor one at a time.

Benefits include:
* easier debugging (if a stressor causes a kernel panic, it's easier
  to determine which stressor was running at the time)
* command line args can be adjusted for each stressor individually
* eliminate redundancy and shorten task run time (some stressors are
  members of multiple classes and thus run 2 or more times)


rasibley:
@cki-bot can you please test ?

rasibley:
@veruu can we please merge ? I see the bot caught one failure for stress-ng which is caused by a hardware flaw on mustangs, after confirming with Jeff.  We could update the test to skip if running on mustangs in the future if this becomes more frequent.

PR #165, opened 2019-04-22
misc/module-load: RHEL8: remove some crypto modules

jbastian:
Crypto modules which are optimized for certain Intel x86_64 processsors will fail on older CPUs and non-x86 CPUs, so remove them from the list.


rasibley:

PR #166, opened 2019-04-23
warn/abort recipe on infrastructure issues

wizardofoz123:
This implements FASTMOVING-591.

When a kernel installation fails, the whole recipe isn't aborted. This
is not the behavior we want. We need warn/abort returned and the job to
be resubmitted.
This patch adds some explicit recipe aborts whenever kernel
installation fails; they seem to be missing by accident.
We also add report-result calls.

This patch will need a followup: we don't really return any special
return codes when we have a failure that's likely an issue of this task
(e.g. script fails to extract kversion from string).

V2
* report warn so skt knows what to do
V3:
* "exit 1" would report fail instead of warn, so report warn, abort recipe,
  then do "exit 0"

### DIFF
--- a/distribution/kpkginstall/runtest.sh
+++ b/distribution/kpkginstall/runtest.sh
@@ -68,8 +68,9 @@ function targz_install()
 
   if [ $? -ne 0 ]; then
     echo "Failed to fetch package from ${KPKG_URL}" | tee -a ${OUTPUTFILE}
+    report_result ${TEST} WARN 99
     rhts-abort -t recipe
-    exit 1
+    exit 0
   fi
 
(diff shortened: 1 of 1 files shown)


rasibley:
@wizardofoz123 here's an example of how we normally set it https://github.com/CKI-project/tests-beaker/blob/master/misc/module-load/runtest.sh#L221 I'm not sure you want 'exit 1' since it may report it as ABORTED/FAIL versus ABORTED/WARN.

veruu:
LGTM, @rasibley could you double check if there isn't anything missed?

rasibley:
LGTM, I don't see any other changes needed.

PR #169, opened 2019-04-24
separate kernel installation and download

wizardofoz123:
This continues work on FASTMOVING-591.

This patch does following:

* If a kernel download fails, it's likely an infrastructure issue, so we
want to report warn/aborted.
* If a kernel package fails to install after being downloaded, it's
  likely a distro issue, so we abort the recipe as well, but report fail
  instead.

### DIFF
--- a/distribution/kpkginstall/runtest.sh
+++ b/distribution/kpkginstall/runtest.sh
@@ -197,6 +197,29 @@ function copr_prepare()
   return 0
 }
 
+function download_install_package()
+{
+  $YUM install --downloadonly -y $1 2>&1 | tee -a ${OUTPUTFILE}
+
+  # If download of a package fails, report warn/abort -> infrastructure issue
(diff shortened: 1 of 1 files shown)


wizardofoz123:
thou shall test this, @cki-bot !

distribution/kpkginstall/runtest.sh
```diff
@@ -197,6 +197,29 @@ function copr_prepare()
   return 0
 }
 
+function download_install_package()
+{
+  $YUM install --downloadonly -y $1 | tee -a ${OUTPUTFILE}
```
veruu: Can we have the stderr redirected here too, or does it cause trouble? It is useful for debugging to have the message logged directly

wizardofoz123: np, added for both download & install

wizardofoz123:
Thanks for review. Merging.

PR #170, opened 2019-04-26
stress-ng: skip the test on Mustang systems

jbastian:
Mustangs have a hardware flaw exposed by stress tests such as stress-ng.  Skip the test if the SUT is a Mustang.


jbastian:
@cki-bot can you please test ?

veruu:
The failure in the pipeline is not related to the PR so I think we can merge this

PR #175, opened 2019-04-30
Add PCI ID and kABI Whitelist symbol removal checks.

ckalina-rh:


ckalina-rh:
Note: Depends on noarch RPM build of `kernel-abi-whitelists`.

ckalina-rh:
As discussed w/ @veruu, the two tests have been moved out of `/misc` to allow for code reuse; target directory: `/legacy`.

PR #176, opened 2019-04-30
WIP: Download RPMs and use downloaded RPMs for install

major:
We currently download some kernel RPMs from the build repository
but then we don't use those downloaded RPMs for the installation.
We also pick and choose which RPMs to install and this makes it
challenging to add new RPMs later.

Instead, download all of the RPMs that were built and then install
them all from the downloaded copy.


major:
Testing this in cloned job 3507285.

major:
This is a WIP -- ignore for now.

major:
@cki-bot test

PR #178, opened 2019-05-02
Rename template/dummy-test to example

spbnick:
Rename a somewhat obscurely-named "template/dummy-test" to "example" to
increase its visibility and inspire its further development.


spbnick:
@rasibley, could you take a look at this?

rasibley:
If running in beaker, at a minimum, the test still needs to source either the rhts or beakerlib environment to be able to communicate back to the harness, best practice is to just use beakerlib:
```
# Include Beaker environment
. /usr/bin/rhts-environment.sh || exit 1
. /usr/share/beakerlib/beakerlib.sh || exit 1
```

However if the purpose is to write a dummy test without the beaker restriction, then looks good to me.

veruu:
@spbnick is this still relevant? Will someone work on the PR or should we close it?

spbnick:
Yeah, would be good to have a more realistic test, even if requiring Beaker. For now though, I'm just increasing its visibility by renaming the existing test.

PR #179, opened 2019-05-03
Onboarding sockopts to CKI test github

ftjahjad:
sanitizing and onboarding  networking sockopts to CKI beaker-test

### DIFF
--- a/networking/sctp/auth/sockopts/Makefile
+++ b/networking/sctp/auth/sockopts/Makefile
@@ -1,0 +1,66 @@
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2015 Red Hat, Inc. All rights reserved.
+#
+#   This copyrighted material is made available to anyone wishing
+#   to use, modify, copy, or redistribute it subject to the terms
+#   and conditions of the GNU General Public License version 2.
+#
+#   This program is distributed in the hope that it will be
+#   useful, but WITHOUT ANY WARRANTY; without even the implied
+#   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#   PURPOSE. See the GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public
+#   License along with this program; if not, write to the Free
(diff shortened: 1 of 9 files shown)


ftjahjad:
@veruu , @rasibley 

Please review this PR

networking/sctp/auth/sockopts/Makefile
```diff
@@ -0,0 +1,66 @@
...
+	@echo "Description:     sctp auth socket ops test" >> $(METADATA)
+	@echo "Type:            functional" >> $(METADATA)
+	@echo "Type:            Singlehost" >> $(METADATA)
+	@echo "TestTime:        60m" >> $(METADATA)
+	@echo "Priority:        Normal" >> $(METADATA)
+	@echo "License:         GPLv2" >> $(METADATA)
+	@echo "Confidential:    no" >> $(METADATA)
+	@echo "Destructive:     no" >> $(METADATA)
+	@echo "RhtsRequires:    kernel-kernel-networking-common" >> $(METADATA)
+	@echo "RhtsRequires:    kernel-kernel-networking-sctp-lib" >> $(METADATA)
```
veruu: Are these some internal test packages? I can't find any matches on a standard Fedora installation. If they are essential for the test can they be ported too? Otherwise specifying how to run the test in the readme is useless because the test will not run properly.

networking/sctp/auth/sockopts/runtest.sh
```diff
@@ -0,0 +1,54 @@
...
+. /mnt/tests/kernel/networking/common/include.sh || exit 1
+
+
+rlJournalStart
+
+########################## Setup ################################
+OUTPUTFILE=$(new_outputfile)
+
+rlPhaseStartSetup
+    rlRun "yum install -y lksctp-tools-devel gcc" 0
```
veruu: While yum seems to still be aliased to dnf on the new systems it might not be in the future. Can we have the test check if it should use yum or dnf, like https://github.com/CKI-project/tests-beaker/blob/master/distribution/kpkginstall/runtest.sh#L163 ?

networking/sctp/auth/sockopts/Makefile
ftjahjad: ported and removed dependencies ..

networking/sctp/auth/sockopts/runtest.sh
ftjahjad: add checking to use dnf if available in the system

PR #180, opened 2019-05-06
Port missing libraries for selinux-policy

idorax:
The shell functions are saved into directory `packages/selinux-policy/serge-testsuite/lib`, and we use ". lib/*.sh" instead of rlImport "selinux-policy/common".


packages/selinux-policy/serge-testsuite/lib/common.sh
```diff
@@ -0,0 +1,268 @@
...
+function rlSEIsStrict() {
+    sestatus | grep -qi strict
+}
+
+
+
+# install requires
+SE_REQUIRES=setools-console
+rlIsRHEL 4 5 && SE_REQUIRES=setools
+rpm -q $SE_REQUIRES || yum -y install $SE_REQUIRES || echo "Error: cannot install $SE_REQUIRES"
```
veruu: While yum is aliased to dnf on the current systems, it may not be the case later. Would it be possible to check whether the system uses yum or dnf and then use that for installation, like for example here: https://github.com/CKI-project/tests-beaker/blob/master/distribution/kpkginstall/runtest.sh#L163 ?

idorax: Hi @veruu, I added a function select_yum_tool(), and updated common.sh accordingly, please take a look again, thanks!

packages/selinux-policy/serge-testsuite/Makefile
```diff
@@ -55,8 +55,47 @@ $(METADATA): Makefile
...
+	@echo "Requires:        libsemanage" >> $(METADATA)
+	@echo "Requires:        expect" >> $(METADATA)
+	@echo "Requires:        lksctp-tools-devel" >> $(METADATA)
+	@echo "Requires:        kernel-modules-extra" >> $(METADATA)
+	@echo "Requires:        kernel-rt-modules-extra" >> $(METADATA)
+	@echo "Requires:        platform-python" >> $(METADATA)
+	@echo "Requires:        python3" >> $(METADATA) >> $(METADATA)
+	@echo "Requires:        python2-lxml" >> $(METADATA) >> $(METADATA)
+	@echo "Requires:        python3-lxml" >> $(METADATA) >> $(METADATA)
+	@echo "Requires:        selinux-policy" >> $(METADATA)
```
rasibley: selinux-policy is listed twice

packages/selinux-policy/serge-testsuite/Makefile
```diff
@@ -55,8 +55,47 @@ $(METADATA): Makefile
...
+	@echo "Requires:        expect" >> $(METADATA)
```
rasibley: expect is listed twice

veruu:
@idorax can you take a look on the duplicates Rachel mentioned? Should be easy to fix. If you don't have time to do it by tomorrow I'll merge this and it can be fixed later as it's not a fatal problem.

idorax:
Hi @rasibley and @veruu , the Makefile is updated and those duplicated packages required are removed. I also removed '/usr/bin/unbuffer' because it is one file of package 'expect'.

PR #184, opened 2019-05-09
Porting network igmp-multicast Conperformance test

ftjahjad:
Porting  igmp-multicast conperformance to github.

### DIFF
--- a/networking/igmp/conformance/common/include.sh
+++ b/networking/igmp/conformance/common/include.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+# variables to control some default action
+NM_CTL=${NM_CTL:-"no"}
+FIREWALL=${FIREWALL:-"no"}
+AVC_CHECK=${AVC_CHECK:-"yes"}
+if [ ! "$JOBID" ]; then
+    RED='\E[1;31m'
+    GRN='\E[1;32m'
+    YEL='\E[1;33m'
+    RES='\E[0m'
+fi
+
+new_outputfile()
+{
+    mktemp /mnt/testarea/tmp.XXXXXX
+}
+
+
+test_pass()
+{
+    #SCORE=${2:-$PASS}
+    echo -e "\n:: [  PASS  ] :: Test '"$1"'" | tee -a $OUTPUTFILE
+    # we don't care how many test passed
+    if [ $JOBID ]; then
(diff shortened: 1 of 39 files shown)


ftjahjad:
Hi @veruu  and @rasibley 

Please review this porting PR.


Thanks.

networking/igmp/conformance/common/include.sh
```diff
@@ -0,0 +1,66 @@
...
+        report_result "${TEST}/$1" "WARN"
+    else
+        echo -e "\n:::::::::::::::::"
+        echo -e ":: [  ${YEL}WARN${RES}  ] :: Test '"${TEST}/$1"'"
+        echo -e ":::::::::::::::::\n"
+    fi
+}
+
+# We only care the main distro
+GetDistroRelease()
```
veruu: Will this work fine on Fedora too? It still returns the distro version and that's only being compared with 7 in the test so I'm guessing it should be fine. Just putting this note out here to keep it in mind in case we can actually run upstream kernels on Fedora and not RHEL in the future, not something that needs to be changed in the test right now :)

networking/igmp/conformance/common/service.sh
```diff
@@ -0,0 +1,101 @@
...
+	local service=$1
+	local action=$2
+	if [ "$(GetDistroRelease)" -ge 7 ];then
+		systemctl $action ${service}.service
+	else
+		service $service $action
+	fi
+}
+
+# TODO: save iptales on RHEL7, then iptables -F
```
veruu: Is the todo needed for proper test functionality?

ftjahjad: I think should be ok to merge without implementing it on my understanding,

@rasibley,  do you have comment and thought?

networking/igmp/conformance/common/include.sh
rasibley: I think it's ok right now since these are workarounds for rhel6

networking/igmp/conformance/common/service.sh
rasibley: Yeah, we see this a lot, the ToDo may no longer be needed or they haven't had time update the test, it runs without issues downstream and should be ok to port as is, as long as it passes continue to pass with upstream kernels :-)

PR #186, opened 2019-05-14
lvm/thinp/sanity: skip test if not enough space

wizardofoz123:
If there's not enough space total space on the device, skip the test.


wizardofoz123:
@cki-bot, test this, bot! Fetch!

veruu:
The pipelines conveniently didn't run on machines where the test would need to be skipped but the change doesn't seem to be breaking anything so merging

PR #188, opened 2019-05-15
set python3-pexpect as a softDependency

rasibley:
@veruu can you please review ? This change sets python3-pexpect as a soft dependency for rhel7, removes unnecessary files, and some cleanup in the library file needed to not conflict with other tests


rasibley:
@veruu all set ! Cleaned up indentation and removed a duplicate line, please review again :-) I also included an updated job in the ticket.

PR #189, opened 2019-05-15
initial commit of ipsec_basic_netns test

rasibley:
@veruu can you please review ?


networking/ipsec/ipsec_basic/ipsec_basic_netns/runtest.sh
```diff
@@ -0,0 +1,208 @@
...
+#
+#	You should have received a copy of the GNU General Public
+#	License along with this program; if not, write to the Free
+#	Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+#	Boston, MA 02110-1301, USA.
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# Global parameters
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+. ../../../common/include.sh
```
veruu: This might be a stupid question but does it work? We had some discussions with Cestmir because he used something similar (include in the runtest but the files were not mentioned in the makefile) and the downloaded test then didn't contain the includes because they were in a different directory. I don't remember if this is the same setup as he has but I thought I'd ask :)

veruu:
The mtools files are being removed in PR 188 already so there will be some conflicts based on which PR will be merged first. Can the removal be removed (hah) from here?

networking/ipsec/ipsec_basic/ipsec_basic_netns/runtest.sh
rasibley: @veruu it does work :-) since we are using restraint, we need to include a metadata file and include repoRequires, otherwise libraries will not be extracted during the fetch from github: https://restraint.readthedocs.io/en/latest/tasks.html#softdependencies
see https://github.com/CKI-project/tests-beaker/pull/189/files#diff-ebc14df65bbe14531bab7b4dc049900bR13

veruu: Oh the reporequires is in metadata! Ok, thanks for explanation!

rasibley:
@veruu  please review again :-) files have been removed with rebase, and passing jobs are in the ticket

networking/ipsec/ipsec_basic/ipsec_basic_netns/ipsec-parameter-setting.sh
```diff
@@ -0,0 +1,169 @@
...
+
+case $IPSEC_PROTO in
+ah)
+	if [ ! -z $AALGO ];then
+		ALG='auth hmac\('$AALGO'\) '$AALGO_KEY
+	elif [ ! -z $ATALGO ];then
+		ALG='auth-trunc hmac\('$ATALGO'\) '$ATALGO_KEY" 96"
+	else
+		echo "Error: ah protocol doesn't set authentication" | tee $IPSEC_PARA_LOG
+		exit 1
```
veruu: If I get correctly how this file is used, it's to set up the passed parameters. Which means if `exit 1` it's a test setup problem and not a kernel problem. Should it abort instead, or am I misinterpreting bash?

rasibley: Good question, the parameters are setup during the rlPhaseStartSetup, if we abort here, we may want to abort for all cases in this file. I will need to get back to you and check with the test maintainer about this.

liuhangbin: As a separate config file, I think the behavior is OK. It will exit 1 and print error message if a user gives the wrong parameter. While in the test, we will give the correct parameter and it should never abort here. BTW, the subtest.yml contains some example SUB_PARAM for upstream users.

veruu: @liuhangbin thanks for the comment! I'd really like to avoid developers complaining about failures because we made a typo in the setup that no one noticed. We require a warn/aborted status for non-kernel issues (this includes failures because the test can't be set up properly) because we can filter that out and not send it to them. Does the reasoning make sense?

If both real issues and setup issues report failure then we need to keep a human in loop to review that problem, and we'd like to avoid that as much as possible as we all have a ton on different responsibilities :)

liuhangbin: OK, makes sense to me. I will ask the author if she want to update it.

Thanks
Hangbin

rasibley: hi @veruu I updated relevant sections in that file to abort if param settings failed, can you review again ?

PR #190, opened 2019-05-16
Port 'kernel/networking/udp/udp_socket' to github (#819)

idorax:
Don't know PACKAGE_NAME is required. Looks it is optional from [1], if it is true, I'll remove it from Makefile, please help to review and confirm.
- [1] https://beaker-project.org/docs/user-guide/task-metadata.html

### DIFF
--- a/networking/udp/udp_socket/Makefile
+++ b/networking/udp/udp_socket/Makefile
@@ -1,0 +1,90 @@
+#
+# Copyright (c) 2019 Red Hat, Inc. All rights reserved.
+#
+# This copyrighted material is made available to anyone wishing
+# to use, modify, copy, or redistribute it subject to the terms
+# and conditions of the GNU General Public License version 2.
+#
+# This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public
+# License along with this program; if not, write to the Free
+# Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+# Boston, MA 02110-1301, USA.
+#
+
+# The toplevel namespace within which the test lives.
+TOPLEVEL_NAMESPACE=kernel
+
+# The name of the package under test:
+PACKAGE_NAME=
+
+# The path of the test below the package:
+RELATIVE_PATH=networking/udp/udp_socket
+
+# Version of the Test. Used with make tag.
+export TESTVERSION=1.1
+
+# The combined namespace of the test.
+export TEST=/$(TOPLEVEL_NAMESPACE)/$(RELATIVE_PATH)
+
+#
+# A phony target is one that is not really the name of a file.  It is just a
+# name for some commands to be executed when you make an explicit request.
+# There are two reasons to use a phony target: to avoid a conflict with a file
+# of the same name, and to improve performance.
+#
+.PHONY: all install download clean
+
+# executables to be built should be added here, they will be generated on the
+# system under test.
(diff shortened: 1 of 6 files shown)


networking/udp/udp_socket/runtest.sh
```diff
@@ -0,0 +1,110 @@
...
+#
+# You should have received a copy of the GNU General Public
+# License along with this program; if not, write to the Free
+# Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+# Boston, MA 02110-1301, USA.
+#
+
+source ../../common/include.sh || exit 1
+
+iproute2_install
```
veruu: Hi @idorax I checked the last Beaker job (3540924) and the taskout log contains

```./runtest: line 22: iproute2_install: command not found```

Can you take a look what happened? There's either a problem with includes or we shouldn't be calling a nonexistent command

File: networking/udp/udp_socket/Makefile [hunk 2]
```diff
@@ -0,0 +1,90 @@
...
+	@echo "Name:         $(TEST)" >> $(METADATA)
+	@echo "TestVersion:  $(TESTVERSION)" >> $(METADATA)
+	@echo "Path:         $(TEST_DIR)" >> $(METADATA)
+	@echo "Description:  udp socket testing ">> $(METADATA)
+	@echo "Type:         Functional" >> $(METADATA)
+	@echo "TestTime:     5m" >> $(METADATA)
+	@echo "RunFor:       kernel" >> $(METADATA)
+	@echo "Requires:     make" >> $(METADATA)
+	@echo "Requires:     gcc" >> $(METADATA)
+	@echo "RepoRequires: networking/common" >> $(METADATA)
```
rasibley: hi @idorax  you need to add repoRequires in a new file 'metadata' otherwise restraint won't fetch the library correctly using 'git fetch'. See https://github.com/CKI-project/tests-beaker/blob/master/networking/route/route_func/metadata#L13 for an example.

networking/udp/udp_socket/runtest.sh
rasibley: Probably because the library was never extracted properly, see my comment above.

idorax: Hi @veruu and @rasibley, `iproute2_install` is a function which is not included in file `networking/common/install.sh` ([1])

- [1] https://github.com/CKI-project/tests-beaker/blob/master/networking/common/install.sh

I removed it because it seems useless when testing RHEL7/8/ark/...
```Bash
# install ip netns support on RHEL6
iproute2_install()
{
	{ [ "$(GetDistroRelease)" -eq 7 ] || ip netns; } && return 0
	local OUTPUTFILE=$(new_outputfile)
	local iproute_ver=${IPROUTE_VER:-"2.6.32-130.el6eng.netns.2"}
	local major_ver=$(echo $iproute_ver | cut -d'-' -f1)
	local minor_ver=$(echo $iproute_ver | cut -d'-' -f2)
	local iproute_netns="http://download-node-02....
       ...<snip>...
}
```

networking/udp/udp_socket/Makefile
idorax: Hi @rasibley, I created two branches to make sure the common files are saved when running case `networking/udp/udp_socket`. It seems that without file `metadata`, it works fine if we add this line to Makefile.
```Makefile
	@echo "RepoRequires: networking/common" >> $(METADATA)
```
For details, please refer to log file http://pastebin.test.redhat.com/764034

Hence, I still use `Makefile` solution having `repoRequires=networking/common` instead of  `metadata` file having `repoRequires=networking/common` to save code lines.

networking/udp/udp_socket/runtest.sh
veruu: Fine by me. We will need to keep it in mind if we start running with RHEL6 but no worries for now.

PR #191, opened 2019-05-16
serge-testsuite: sync with downstream

WOnder93:
This in particular fixes an issue with SCTP support and adds other minor
improvements.


rasibley:
Thank you @WOnder93 ! I just ran a few test jobs and all green, @veruu I'll link them in the ticket, can you please review ?

packages/selinux-policy/serge-testsuite/runtest.sh
```diff
@@ -66,6 +75,17 @@ rlJournalStart
...
 
+        # version_le() sanity check:
+        rlRun "version_le 4.10 4.10"
+        rlRun "version_le 4.10 4.10.0"
+        rlRun "version_le 4.10 4.10.1"
+        rlRun "! version_le 4.10 4.9"
+        rlRun "! version_le 4.10.0 4.10"
+
+        # make sure the right version of kernel[-rt]-modules-extra is installed
+        rlRun "yum install -y kernel-modules-extra-$(uname -r)" 0-255
```
veruu: Can you please pick dnf/yum and use that, or simply use https://github.com/CKI-project/tests-beaker/blob/master/packages/selinux-policy/serge-testsuite/lib/common.sh#L264 to make sure the test won't need to be fixed once we run on Fedora that won't have dnf symlinked to yum?

WOnder93: You're right, I didn't realize I can't rely on the `yum` command being always available...

I have now corrected this downstream. I didn't use `select_yum_tool` in the end, because it doesn't seem to be exported by the library downstream. I'll force-push an updated version in a while.

PR #196, opened 2019-05-17
blktests: disable block/017 on s390x

yizhanglinux:
Signed-off-by: Yi Zhang <yi.zhang@redhat.com>


rasibley:
Thanks Yi, @veruu  can you please review ? blktests is failing on s390x, we need to disable while we investigate the root cause.

PR #200, opened 2019-05-20
legacy: pci id test changes

ckalina-rh:
The first change in the PR fixed kmod search path (as there are debug modules present as well, this was no longer unique).

The second one relaxes the stringent requirement that the set of PCI IDs be equal on a string basis to the requirement that they be syntactically equivalent up to shell pattern expansion (cf. commit message for an example).

Signed-off-by: Cestmir Kalina <ckalina@redhat.com>


ckalina-rh:
Force pushed to fix typo in the commit message.

PR #204, opened 2019-05-22
initial commit of vxlan basic test

rasibley:
@veruu please review :-)

### DIFF
--- a/networking/tunnel/vxlan/basic/Makefile
+++ b/networking/tunnel/vxlan/basic/Makefile
@@ -1,0 +1,69 @@
+# Copyright (c) 2015 Red Hat, Inc. All rights reserved. This copyrighted material 
+# is made available to anyone wishing to use, modify, copy, or
+# redistribute it subject to the terms and conditions of the GNU General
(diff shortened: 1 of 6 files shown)


rasibley:
hi @veruu this isn't a multi-host test, we currently run this on single hosts now, just porting so we can run on upstream kernels. If we were to enable multi-host we would need to confirm the hosts were available to us or send over the UMB.

veruu:
@rasibley hm, ok. The test says `--topo=multihost` and deals with client/server configuration so it's confusing. Can you mention it can run on a single host in the test please to make it clear?

rasibley:
@veruu well there is two ways to runs it, singlehost or multihost:
 attr: {level: Tier1, type: singlehost, distronotin: RHEL-6}

Would it be ok if we left it in case we can run on a multi-host environment over umb in the future ? Otherwise I can remove it.

veruu:
@rasibley sure, we can keep the functionality for future cases and just putting "can run standalone or multihost" into the readme file would be enough

PR #206, opened 2019-05-24
skip test if unable to install perf related packages

rasibley:
@veruu  please review, I'll include beaker jobs in the ticket


rasibley:
@spbnick is looking into setting this to brew builds only, we can close this PR

PR #212, opened 2019-05-29
add mdadm case trim and include

zhongchanghui:
add storage/mdadm/include and trim for CKI, please have a look and merged if it ok for you.


storage/mdadm/include/PURPOSE
```diff
@@ -0,0 +1,6 @@
+Please check http://wiki.test.redhat.com/fge/mdadm_include
```
veruu: This page doesn't seem to exist when I tried to access it. Can instead the information needed (unless including customer data) be added to this file instead of being linked too?

storage/mdadm/include/include.sh
```diff
@@ -0,0 +1,619 @@
...
+if [ "$info" = "write-journal" ]; then
+        JOURNAL_SUPPORT=1
+fi
+
+#Install fio from upstream
+function install_fio() {
+
+	git_url=git://git.kernel.org/pub/scm/linux/kernel/git/axboe/fio.git
+
+	yum install libaio-devel zlib-devel -y
```
veruu: Can yum/dnf be chosen based on which one is available instead of blind yum use? Multiple other tests are already doing this check if you grep the repo so you can just reuse that.

storage/mdadm/include/include.sh
```diff
@@ -0,0 +1,619 @@
...
+	git clone $git_url
+	echo "INFO: Installing Fio"
+	cd fio  &&./configure && make && make install
+	which fio
+	if [ $? -ne 0 ]; then
+		echo "FAIL: Fio not succesffully installed"
+		exit 1
```
veruu: Is this a test prerequisite or is the test also testing the ability to install fio? If it's just a prerequisite we don't want to report this failure to developers and the test should abort instead (see the top level readme for details). I don't see the function being used in this PR though.

storage/mdadm/include/include.sh
```diff
@@ -0,0 +1,619 @@
...
+#       RETURN_STR		# $md_raid like '/dev/md0'
+#		MD_DEVS			# $raid_dev_list like '/dev/sda /dev/sdb'
+#----------------------------------------------------------------------------#
+
+function MD_Create_RAID (){
+	EX_USAGE=64 # Bad arg format
+	if [ $# -lt 3 ]; then
+		echo 'Usage: MD_Create_RAID $level $dev_list $raid_dev_num \
+		[$spar_dev_num] [$chunk]'
+		exit "${EX_USAGE}"
```
veruu: This looks like a test / beaker XML configuration problem, rather then a real kernel failure. So I have the same question as above -- we don't want to report this to developers.

storage/mdadm/include/include.sh
```diff
@@ -0,0 +1,619 @@
...
+	local ret=0
+	# start to create
+	echo "INFO: Executing MD_Create_RAID() to create raid $level"
+	# check if the given disks are more the needed
+	for i in $dev_list; do
+		dev_num=$((dev_num+1))
+	done
+	if [ $dev_num -lt $(($raid_dev_num+$spar_dev_num)) ]; then
+		echo "FAIL: Required devices are more than given."
+		exit 1
```
veruu: Same here.

Actually also everywhere below with nonzero exits. This is just a test include so it all depends on the interpretation of the exit code in the test itself

storage/mdadm/trim/Makefile
```diff
@@ -0,0 +1,85 @@
...
+	@echo "TestVersion:  $(TESTVERSION)"    >> $(METADATA)
+	@echo "Description:  test the function trim support">> $(METADATA)
+	@echo "TestTime:     2h" >> $(METADATA)
+	@echo "RunFor:       $(PACKAGE_NAME)" >> $(METADATA)
+# add any other packages for which your test ought to run here
+	@echo "Requires:     $(PACKAGE_NAME)" >> $(METADATA)
+	@echo "Requires:	 mdadm" >> $(METADATA)
+	@echo "Requires:         iscsi-initiator-utils" >> $(METADATA)
+	@echo "Requires:         scsi-target-utils" >> $(METADATA)
+	@echo "RhtsRequires: tests-beaker-storage-mdadm-include" >> $(METADATA)
```
veruu: I think @rasibley mentioned using `RepoRequires` for libraries. I'm not sure if this way is going to work since the include is not in the RH Beaker task library.

storage/mdadm/include/PURPOSE
rasibley: Also we don't want to include internal hostnames in open sourced tests :-)

storage/mdadm/include/PURPOSE
```diff
@@ -0,0 +1,6 @@
...
+
+Add these lines into your Makefile:
+===
+@echo "RhtsRequires: rh-tests-kernel-storage-include" >> $(METADATA)
+@echo "RhtsRequires: rh-tests-kernel-storage-mdadm-include" >> $(METADATA)
```
rasibley: These libraries need to be open sourced, see the networking tests as an example including the metadata file which extracts the library, I would just remove this section completely: https://github.com/CKI-project/tests-beaker/blob/master/networking/socket/fuzz/metadata

veruu: Since I don't see the lines in metadata in the PR them may not be needed at all?

storage/mdadm/include/include.sh
rasibley: Here's an example https://github.com/CKI-project/tests-beaker/blob/master/networking/sctp/auth/sockopts/runtest.sh#L30

storage/mdadm/trim/Makefile
rasibley: Yes will need to open source this library as well so it's self contained, take a look at the networking tests and example of the metadata I mention above

veruu: I thought it tries to point to the included library from the other files in this PR (the mdadm/include files)

storage/mdadm/include/include.sh
rasibley: Look like this would qualify as a setup/infra failure, we need to abort the recipe in these cases, see https://github.com/CKI-project/tests-beaker/blob/master/acpi/acpitable/runtest.sh#L34

storage/mdadm/trim/Makefile
rasibley: ah your right, I see it now, so just need to reference the include directory versus the beaker task, I think you can either do it in the Makefile using RepoRequires or metadata (restraint) using repoRequires, there are lots of examples in the repo.

storage/mdadm/include/PURPOSE
zhongchanghui: I have delete unnecessary files from common (include renamed common)

storage/mdadm/include/include.sh
zhongchanghui: Remove the feature definitions that are not needed temporarily (include.sh is renamed to libmdadm.sh), and will follow the provided examples when needed later.

storage/mdadm/include/include.sh
zhongchanghui: If the instruction fail to execute, just use "exit", or don't use "exit" to return its own information and exit code. Which method is appropriate?

storage/mdadm/include/PURPOSE
zhongchanghui: Yes, don't need this file, delete it

zhongchanghui:
https://beaker.engineering.redhat.com/jobs/3577215
Can you take a look the test results on the beaker? If it meets the require? Please feel free to contact me.

storage/mdadm/include/include.sh
veruu: Depends on how the result is interpreted. This file was removed but if you check the exit code in the test then you can exit with code.
When checking the exit code, the rule is: fail if it's a kernel bug, abort if it is not.

Does this answer your question? I think the current code looks ok

veruu:
The job looks fine to me, thanks! I resubmitted it without the "common" task as it's only a library and not a task that should be executed to verify it works fine. @rasibley do you have any other comments I missed or can we merge it after I see the results from my job?

rasibley:
@veruu that was only question related to the common task, glad it works just fine w/o it, thanks again @zhongchanghui for your PR, no more comments from me :-)

veruu:
None of my jobs seems to get past the installation so I'm just going to merge it now and we can clean out the details with the kpet-db onboarding

PR #217, opened 2019-06-03
port tracepoint operational test

ftjahjad:
Port Systemtap Tracepoint Operational Test

### DIFF
--- a/tracepoints/operational/runtest.sh
+++ b/tracepoints/operational/runtest.sh
@@ -1,0 +1,232 @@
+#!/bin/bash
+
+# Source the common test script helpers
+. /usr/bin/rhts-environment.sh || exit 1
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+
+
+STAP_VERBOSE_FLAG="-v"
+if [ x"${STP_VERBOSE}" = x"y" ]; then
+    STAP_VERBOSE_FLAG="-vvvv"
+fi
+
+# Helper functions
+function resultFail()
+{
+    echo "***** End of runtest.sh *****" | tee -a $OUTPUTFILE
+    report_result $1 FAIL $2
+    echo "" | tee -a $OUTPUTFILE
+}
+
+function resultPass ()
+{
+    echo "***** End of runtest.sh *****" | tee -a $OUTPUTFILE
+    report_result $1 PASS $2
+    echo "" | tee -a $OUTPUTFILE
+}
+
+function submitLog ()
+{
+    LOG=$1
+    if [ -z "$TESTPATH" ]; then
+        echo "Running in developer mode"
+    else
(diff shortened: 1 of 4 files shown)


ftjahjad:
Hi @veruu , and @rasibley 

Please review the PR.

Thanks,

tracepoints/operational/runtest.sh
```diff
@@ -0,0 +1,232 @@
...
+            rm -f $PROBES_NAME_FILE
+        fi
+    done
+}
+
+function runTest ()
+{
+    local TESTLIST=`/usr/bin/stap -L 'kernel.trace("*")' | grep -o "\".*\""`
+    if [ -z "$TESTLIST" ] ; then
+        resultFail TESTLIST_EMPTY 99
```
veruu: I'm not sure if this is something we should report as a failure or a prerequisite for the testing (in which case we should skip or abort), as I'm not familiar with this at all. @ftjahjad can you check which one it is?

tracepoints/operational/runtest.sh
```diff
@@ -0,0 +1,232 @@
...
+
+# Checking if running kernel-rt
+if [ $(is_kernel_rt) -eq 0 ]; then
+    ${yum} -y install kernel-rt-devel-${kernel}
+else
+    ${yum} -y install kernel-devel-${kernel}
+fi
+res=$?
+if [ $res -ne 0 ] ; then
+    echo "Failed to install kernel devel-${kernel}" | tee -a $OUTPUTFILE
```
veruu: What will the exit code be here? Should an explicit warn be reported?

tracepoints/operational/runtest.sh
jbastian: If SystemTap cannot find any kernel tracepoints, I would consider that a failure.  Unless you intend to build a kernel with tracepoints disabled, this should be a failure (not an abort).

ftjahjad: @jbastian , thanks. it is clear now.

tracepoints/operational/runtest.sh
ftjahjad: adding "report  warn" before aborting.

ftjahjad:
@veruu 
Thanks for pointing put the missing "warn/aborted"  report. :+1: 

Please  review  changes to report   "warn/aborted"  before aborting the recipe.

tracepoints/operational/runtest.sh
veruu: We either build normal tarballs (upstream) or use rpmbuild with the following options:

```--with=kabidw_base --without=bpftool --without=debug --without=doc --without=ipaclones --without=perf --without=selftests --without=tools
```
The configs used are the standard configs used with the release.
So I think the building part should include tracepoints and everything should be ok. Thanks @jbastian !

PR #220, opened 2019-06-05
Update the test case because system hit panic when testing RHEL7

zhongchanghui:
Since rhel7's support for MD write-journal RAID is unstable, temporarily delete the journal raid part.
please take a look.

beaker job id: 6963820

### DIFF
--- a/storage/swraid/trim/PURPOSE
+++ b/storage/swraid/trim/PURPOSE
@@ -1,2 +1,2 @@
 PURPOSE of trim
-Description: Test the function of create raid0 1 4 5 6 10 and write-journal raid 4 5 6 with mdadm, format and mount test.
+Description: Test the function of TRIM which quickly erase invalid data for the RAID that created by mdadm.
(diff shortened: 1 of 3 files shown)


rasibley:
hi @zhongchanghui there are conflicts, I think your fork wasn't synced with master before you pushed your changes, can you please rebase ? Also does this fail only on rhel7 ? Why not instead add a condition to disable that part when running on rhel7, this way we don't have to disable the entire section.

zhongchanghui:
Hello, Rachel, I have solved the conflict. The function of write-journal in mdadm is enabled on rhel7, but it is disabled in rhel8 now.I want to ensure the stability of md CKI case, so delete it temporarily. This part of the code, in the future can be added in the new case or the existing case to enable the write-journal function.

zhongchanghui:
@rasibley

rasibley:
Thank you @zhongchanghui  I hope we can re-enable that part after some time, I'm ok with the following changes, we can retest in kpet-db, @veruu is on PTO, but maybe @major can approve ?

PR #223, opened 2019-06-07
module-load: remove ppp_mppe since it fails in FIPS mode

jbastian:


rasibley:
@major can you review ?

PR #226, opened 2019-06-10
Enhance case 'ipsec_basic_netns' to avoid random failure

idorax:


idorax:
Hi @rasibley, I removed those two lines related to rhel6 only, a beaker job `3601148` was started to test it on x86_64. Please take a look again, thanks!

PR #227, opened 2019-06-11
Fix vpd test in case it fails to get test device

guazhangRH:


rasibley:
Thank you @guazhangRH looks good to me, @veruu can you please review ? This will help stabilize the vpd test.

PR #228, opened 2019-06-11
Fix KVM Kernel Self Tests

mcondotta:
Make Kernel Self Tests run with the current running Kernel.
Remove patch since the code has been added on RHEL8 (4.18.0-97) and ARK (5.2).
Fix KVM Kernel Selftests to run for the correct kernel version (currently running).
Run indiviual tests separately to be able to disable them if there are bugs.
Fix the script to detect failures.
Fix python for RHEL8.1.
Fix downloading of the SRC RPM for CKI builds.

Test Jobs:
RHEL8.1 Intel x86_64: https://beaker.engineering.redhat.com/jobs/3600494
RHEL8.1 AMD x86_64: https://beaker.engineering.redhat.com/jobs/3600492
ARK Intel x86_64: https://beaker.engineering.redhat.com/jobs/3600504
ARK AMD x86_64: https://beaker.engineering.redhat.com/jobs/3600502


vm/kvm-self-tests/runtest.sh
```diff
@@ -141,13 +217,29 @@ function setup
...
 
+    # if running on rhel8, use python3
+    if grep --quiet "release 8." /etc/redhat-release && [ ! -x /usr/bin/python ];then
+        ln -s /usr/libexec/platform-python /usr/bin/python
+    fi
+
     rlRun "cd $TMPDIR"
-    rlFetchSrcForInstalled $pkg
-    typeset rpmfile=$(ls -1 $TMPDIR/${pkg}*.src.rpm)
+    dnf download ${pkg} --source
```
veruu: Is dnf present on older releases? In other tests, we check if it's yum or dnf that's present and use that for installations

mcondotta: Great suggestion, this test is only running for RHEL8 and ARK, however I think it's a good idea to test before running the command. I'll copy the check from kpkginstall runtest.sh and try both DNF and YUM before falling back to the beakerlib.

PR #230, opened 2019-06-13
Add iscsi/params test (#827)

idorax:
`iscsi/params` test case is for testing targetcli by leveraging `https://gitlab.com/rh-kernel-stqe/python-stqe`, please help to review, thanks!

### DIFF
--- a/storage/include/libstqe.sh
+++ b/storage/include/libstqe.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Red Hat, Inc. All rights reserved.
+#
+# This copyrighted material is made available to anyone wishing
+# to use, modify, copy, or redistribute it subject to the terms
+# and conditions of the GNU General Public License version 2.
+#
+# This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public
+# License along with this program; if not, write to the Free
+# Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+# Boston, MA 02110-1301, USA.
+#
+
+source /usr/share/beakerlib/beakerlib.sh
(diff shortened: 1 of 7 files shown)


idorax:
Test result for x86 looks good, please refer to bkr job `3606649`.

storage/include/libstqe.sh
```diff
@@ -0,0 +1,57 @@
...
+    # clone the framework
+    typeset fwroot=$(stqe_get_fwroot)
+    rlRun "rm -rf $fwroot"
+    rlRun "git clone $STQE_GIT $fwroot" || return 1
+
+    # install the framework
+    pushd "." && rlRun "cd $fwroot"
+
+    [[ -n $STQE_STABLE_VERSION ]] && \
+        rlRun "git checkout $STQE_STABLE_VERSION" || return 1
```
veruu: Can we have the setup failures here and below abort the test? We don't want to report failures to developers because the repo couldn't be cloned, for example.

idorax: Hi @veruu, for the test log having a failure, I'll mail to the test maintainer and copy to you soon.

For the setup failures, I updated `storage/iscsi/params` and let it go to rlSkip(). Two beaker jobs are done, one (`3614622`) is for testing the latest test code, the other (`3614634`) is for testing it does go to rlSkip() by injecting `return 2` to function `stqe_init_fwroot()`. Please take a look again, thanks!

veruu: Thanks for the changes and looking into the issues! However I don't think the skipping is right to do here - we usually skip tests when the HW is not supported by a test and no one looks twice on it. But if eg. the git clone fails it may mean the repo was removed and we need to fix it. Aborted jobs grab people's attention as they clearly mark something is wrong.

idorax: Hi @veruu, the test code is updated according to your comments. And I tested it via beaker.

- Abort: `3617291`
- Pass: `3617293`

BTW, Filip helped to fix `FAIL:(libsan.host.mp) ...` failure, now the test log looks fine.

veruu: Looking good now, thanks @idorax and Filip for fixing the issues!

PR #231, opened 2019-06-14
distribution/kernel-fips-mode: add test for kernel FIPS mode

The-Mule:
Test for kernel FIPS mode. It is applicable for RHEL-8 only and it might be potentially destructive. Any kernel FIPS issue might cause kernel panic and break subsequent testing. Therefore it is important to run this test either in separation or as the last test.


The-Mule:
@cki-bot test

veruu:
@The-Mule even if you had permissions to trigger testing, the test needs to be onboarded in kpet-db to actually run it. I don't see any obvious problems in the test itself so I'm going to merge this and then you can work with @ftjahjad and @spbnick on the kpet-db changes, as Rachel is on PTO.

PR #234, opened 2019-06-19
legacy: Use nightly repo for pci id/kabi-whitelist checks, add RHEL7-specific package for pci id comparison

ckalina-rh:
- Nightly repository is required for proper baseline comparison.
 - Add `yum-utils` requires -- use yumdownloader instead of `(yum|dnf) ([re]install|update) --downloadonly` -- simplifies RPM download code
 - Also add kernel package as a source for kernel modules for RHEL7.

### DIFF
--- a/legacy/shared/rpm-utils.sh
+++ b/legacy/shared/rpm-utils.sh
@@ -175,7 +175,8 @@ function rpm_install()
 
 function rpm_extract()
 {
-        __yum_call --downloadonly --downloaddir=$RPM_TMPDIR "${RPM_EXTRACT[@]}"
+        yumdownloader --disablerepo='*' --enablerepo=rhel-latest \
+                      --downloaddir=$RPM_TMPDIR "${RPM_EXTRACT[@]}"
 
         if test $? -gt 0
         then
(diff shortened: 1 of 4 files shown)


legacy/shared/rpm-utils.sh
```diff
@@ -175,7 +175,8 @@ function rpm_install()
 
 function rpm_extract()
 {
-        __yum_call --downloadonly --downloaddir=$RPM_TMPDIR "${RPM_EXTRACT[@]}"
+        yumdownloader --enablerepo=latest --disablerepo='kernel-cki' \
```
spbnick: I'm not sure if we have the `kernel-cki` name documented/guaranteed anywhere, so I'd rather just disable all other repos. Like this:

    yumdownloader --disablerepo='*' --enablerepo=latest \
                  --downloaddir=$RPM_TMPDIR "${RPM_EXTRACT[@]}"

ckalina-rh: @spbnick Good catch, resolved.

spbnick:
LGTM!

spbnick:
Argh, I forgot to ask for a repo name change here ("latest" -> "rhel-latest"). So, if we do it, you'll need another change.

ckalina-rh:
> Argh, I forgot to ask for a repo name change here ("latest" -> "rhel-latest"). So, if we do it, you'll need another change.

Already done that.

spbnick:
Right. I'm blind :)

PR #236, opened 2019-06-19
misc/kernel-fips-mode: move from distribution test bucket to misc

The-Mule:


The-Mule:
I noticed that distribution bucket was probably not the best place for the test. As it does not fit into any specific category, I put it into misc.

The-Mule:
Wow, such a quick merge, thanks @veruu!

veruu:
It's just a file movement so nothing to think about, and I just happened to look at my emails as you submitted the pull so great timing!

PR #237, opened 2019-06-26
Add iscsi/params test case (#827) (cont)

idorax:
For RHEL7, use python2 instead because python3 is not available by default


idorax:
Two beaker jobs are started to test the fix.

- RHEL7: `3630551`
- RHEL8: `3630542`

PR #241, opened 2019-07-09
fs/xfs/xfstests: do not report failure if subcase does not exist

jencce:
Remove 2 testcases that has been renamed upstream.

Signed-off-by: Murphy Zhou <xzhou@redhat.com>

### DIFF
--- a/filesystems/xfs/xfstests/RUNTESTS
+++ b/filesystems/xfs/xfstests/RUNTESTS
@@ -217,5 +217,3 @@ generic/508
 generic/509
 generic/510
 generic/511
-shared/001
-shared/003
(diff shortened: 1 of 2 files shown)


veruu:
If the testcases were just renamed, wouldn't it make more sense to add those into the list instead of just removing the old names?

rasibley:
Thank you @jencce looks good, @veruu can you review ?

veruu:
@rasibley I already left a question here in comments

rasibley:
oops if I had looked ^^ I would have noticed, thanks @veruu !

jencce:
shared/001 renamed to ext4/036
shared/003 renamed to ext4/037

Our listed cases are running for both xfs and ext4. These 2 cases only run on ext4, they are skipped when running on xfs. So I removed them.

veruu:
Thanks for the explanation @jencce ! I'll add the info to the commit message if I'm able to do so while merging the change, please include things like this in there in the future so it's clear what happened

jencce:
Thanks very much @veruu ! I should include more info in the commit message. Feel free to let me redo the commit if needed next time.

PR #242, opened 2019-07-09
Enable CPU Idle Test (x86_64 only) test to kpet-db (#982)

idorax:


rasibley:
@idorax do you have some example test jobs ? Also David mentions there is some hand waving involved, what does that mean exactly ? Was the test already updated to account for that ? Where can we run this test in beaker ?

idorax:
Hi @rasibley, a beaker job is done. Please take a look at `3668037`. For your questions, let me check with David.

rasibley:
Thanks for the test job @idorax ! @veruu can you review ?

cpu/idle/runtest.sh
```diff
@@ -0,0 +1,68 @@
...
+        return $BKRM_FATAL
+    fi
+
+    # check this test is supported by CPU
+    rlRun "rdmsr 0x606" || rlSkip "su access is not available"
+    if (( $? != 0 )); then
+        rlSetReason $BKRM_UNSUPPORTED \
+            "su access is not available"
+        return $BKRM_UNSUPPORTED
+    fi
```
veruu: This looks like it would report skip twice (once for line 50 and once for 54) but that shouldn't create any problems, IMO.

rasibley: @veruu yeah that's how I see it as well, and I agree with you shouldn't cause an issue since only the first would be invoked.

PR #246, opened 2019-07-10
kdump-sysrq-c: Analyze a vmcore by crash utilities

emmawu:
Added steps to validate a vmcore captured after a sysrq-c panic by crash utitlies.
To run crash against a vmcore, it requires to install crash and kernel-debuginfo/kernel-debuginfo-common packages. kernel-debuginfo/kernel-debuginfo-common will be installed from kernel-cki repo.

### DIFF
--- a/kdump/include/runtest.sh
+++ b/kdump/include/runtest.sh
@@ -299,7 +299,7 @@ GetBiosInfo()
 GetHWInfo()
 {
     Log "- Getting system hw or firmware config."
-    rpm -q lshw || yum install -y lshw
+    rpm -q lshw || InstallPackages lshw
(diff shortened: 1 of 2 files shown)


veruu:
Thanks for the PR! However, could you please detect whether it's yum or dnf on the system and use that instead of hardcoded yum? You can take inspiration from various other tests here that do so already, for example here: https://github.com/CKI-project/tests-beaker/blob/master/networking/common/install.sh#L8

veruu:
@cki-bot can you test this in the meanwhile please?

veruu:
Okay, we can't test this now because the baselines kernels were deleted...

PR #247, opened 2019-07-12
kdump-sysrq-c: Check kernel debuginfo vmlinux file

emmawu:
kernel-debuginfo package cannot be installed from a yum source if the kernel is installed from a tar.gz.
In this case, the "/usr/lib/debug/lib/modules/$(uname -r)/vmlinux" is supposed to be taken care by cki boot test. Then kdump-sysrq-c will only check if the file exists.

### DIFF
--- a/kdump/include/runtest.sh
+++ b/kdump/include/runtest.sh
@@ -205,8 +205,10 @@ LogRun() {
 
 Skip() {
     local msg="$1"
-    echo "- warn: $msg" | tee -a "${OUTPUTFILE}"
+    echo "- skip: $msg" | tee -a "${OUTPUTFILE}"
     skip=$((skip + 1))
+
+    Report
 }
 
 Warn() {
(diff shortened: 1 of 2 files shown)


emmawu:
Currently if kernel is installed from a tarball, it won't provide the "/usr/lib/debug/lib/modules/$(uname -r)/vmlinux" file which contains kernel debuginfo.  This change will not solve the problem of missing debuginfo vmlinux file, but only provide a proper error log.  

You may workaround this error by passing following param to kdump sysrq-c test:
     <param name="ANALYZE_VMCORE" value="false"/>
Then kdump sysrq-c test will not try to install kernel debuginfo for analyzing the vmcore.

veruu:
It looks to me like a missing debuginfo is a setup/support problem and not a kernel bug. If we just report a failure then the developers will be sent an email which basically says they have a bug in their patches, which is not the case here. Could we rather report a skip or abort in this case?

@rasibley should be setting up the parameter for upstream kernels so we should stop seeing the problem and if it reappears, it indicates we need to fix something

rasibley:
@veruu I agree, @emmawu I added the param to kpet-db, can we just skip or abort in this case since it sounds like a setup problem ?

emmawu:
@rasibley 
Done,  updated. It will skip steps to analyse a vmcore if no kernel-debuginfo vmlinux is found and kernel is installed from a tarball.
The test will report 'PASS' but show logs in test:

> \- Kernel is installed from a tar, not from yum/dnf package.
> \- kernel-debuginfo should be prepared in cki boot test.
> \- Check if /usr/lib/debug/lib/modules/5.2.1-rc1-61731e8.cki/vmlinux exists
> \- Failed to find /usr/lib/debug/lib/modules/5.2.1-rc1-61731e8.cki/vmlinux.
> \- Skip running crash utitlies against the vmcore.

emmawu:
@rasibley , I may mis-understood... Do you like the test to report a SKIP (or warn)? or skip part of test but report a PASS?  It might be confusing to report a SKIP because it doesn't skip the entire test but only part of steps.

rasibley:
@emmawu I think the point is to skip that section where it analyzes the vmcore, we don't want to report it as a WARN or FAIL if that code is skipped, since it will be flagged as a failure in beaker and CKI would report it as an overall FAIL. If the rest of the test can run without the analysis, then skipping just part of the test is fine, but it would have to skip and PASS since it's not supported.

emmawu:
@rasibley 
I see..make sense. Fixed the test script a bit again. It would skip the part that analyzing the vmcore if no debug vmlinux is found no matter how kernel is installedl. It would still report a PASS if rest of test passed.

PR #251, opened 2019-07-16
perf: adding a new testsuite (wrapper)

frozenstein:
The perftool-testsuite should become a part of CKI perf testing as well.
It provides a different approach of perf testing than the internal-testsuite.
The wrapper is based on the wrapper used at Red Hat, but adjusted for the
CKI usage. It contains a blacklist to mask out known issues and unstable
tests, since CI should be stable.

### DIFF
--- a/packages/perf/perftool-testsuite/runtest.sh
+++ b/packages/perf/perftool-testsuite/runtest.sh
@@ -1,0 +1,293 @@
+#!/bin/bash
+# vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   runtest.sh of /tools/perf/Sanity/perftool-testsuite-cki
+#   Description: the test runs upstream perftool-testsuite in CKI
+#   Author: Michael Petlan <mpetlan@redhat.com>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2017 Red Hat, Inc.
+#
+#   This program is free software: you can redistribute it and/or
+#   modify it under the terms of the GNU General Public License as
+#   published by the Free Software Foundation, either version 2 of
+#   the License, or (at your option) any later version.
+#
+#   This program is distributed in the hope that it will be
+#   useful, but WITHOUT ANY WARRANTY; without even the implied
+#   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#   PURPOSE.  See the GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public License
+#   along with this program. If not, see http://www.gnu.org/licenses/.
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+# Include Beaker environment
+. /usr/bin/rhts-environment.sh || exit 1
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+. blacklist.sh
+
+PACKAGE="perf"
+
+# configuration
+GIT_CLONE_ATTEMPTS_COUNT=25
+KEEP_LOGS="no"
+PERFTESTS_ENABLE_BLACKLIST=${PERFTESTS_ENABLE_BLACKLIST:-1}
+
+# hook, someone likes using "True" there, we like 1, 0 values more
+if [ "$PERFTESTS_ENABLE_BLACKLIST" = "true" -o "$PERFTESTS_ENABLE_BLACKLIST" = "True" ]; then
+	PERFTESTS_ENABLE_BLACKLIST=1
+fi
+
+skip_testcase()
+{
+	echo "$1" | tee -a ${OUTPUTFILE}
+	rhts-report-result $TEST SKIP $OUTPUTFILE
+	exit 0
+}
+
+
+fetch_the_testsuite()
+{
+	test -e perftool-testsuite && return
+
+	REPO_CLONED=0
+	for (( i=0; i<$GIT_CLONE_ATTEMPTS_COUNT; i++ )); do
+		echo "Trying to clone the repo....... take $i..."
+		git clone https://github.com/rfmvh/perftool-testsuite.git
+		if [ $? -eq 0 ]; then
+			echo "Oh dear, I have been able to clone the repo. Yeah!!!"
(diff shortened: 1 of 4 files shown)


packages/perf/perftool-testsuite/runtest.sh
```diff
@@ -0,0 +1,293 @@
...
+
+		rlLog "Variables:"
+		rlLog "KERNEL = $KERNEL"
+		rlLog "KERNEL_PKG_NAME = $KERNEL_PKG_NAME"
+		rlLog "KERNEL_DEBUGINFO_PKG_NAME = $KERNEL_DEBUGINFO_PKG_NAME"
+		rpmquery $KERNEL_DEBUGINFO_PKG_NAME
+		if [ $? -ne 0 ]; then
+			# we need to install debuginfo for the proper kernel
+			# but sometimes, debuginfo-install is not available!
+			which debuginfo-install || rlRun "yum -y install yum-utils dnf-utils" 0 "Installing {yum,dnf}-utils (it has not been present)"
```
veruu: Could you please detect whether dnf or yum is available and use that instead of hardcoding yum? You can use similar code as the other tests are, e.g. https://github.com/CKI-project/tests-beaker/blob/master/networking/common/install.sh#L8

packages/perf/perftool-testsuite/runtest.sh
```diff
@@ -0,0 +1,293 @@
...
+
+		# set kptr_restrict to 0
+		PREVIOUS_KPTR_RESTRICT=`cat /proc/sys/kernel/kptr_restrict`
+		rlRun "echo 0 > /proc/sys/kernel/kptr_restrict"
+		rlAssert0 "kptr_restrict must be set to 0" `cat /proc/sys/kernel/kptr_restrict`
+
+		# clone the upstream perftool-testsuite repo
+		fetch_the_testsuite
+		# return SKip if the testsuite could not be fetched
+		test -d "perftool-testsuite" || skip_testcase "Could not fetch the upstream testsuite from github. I am sorry, dude."
```
veruu: Would it make more sense to abort the execution if the upstream testsuite can't be reached? Aborted tests grab attention and not being able to clone the repo might also mean that the repo was migrated or removed (extreme, but might happen) and we won't know about it as people don't usually look at the logs of skipped tests.

frozenstein:
On Tue, 16 Jul 2019, Veronika Kabatova wrote:
> 
> @veruu requested changes on this pull request.
> 
> Thanks for the PR! I have just two comments from the CKI perspective (one about the test result interface we use, one to ensure the test will work with new distros well without changes). Can
> you please take a look when you have time?
> 
> ______________________________________________________________________________________________________________________________________________________________________________________________
> 
> In packages/perf/perftool-testsuite/runtest.sh:
> 
> > +			export KERNEL_DEBUGINFO_PKG_NAME="kernel-debug-debuginfo-$KERNEL"
> +			if [ $(is_kernel_rt) -eq 0 ]; then
> +				export KERNEL_PKG_NAME="kernel-rt-debug-$KERNEL"
> +				export KERNEL_DEBUGINFO_PKG_NAME="kernel-rt-debug-debuginfo-$KERNEL"
> +			fi
> +		fi
> +
> +		rlLog "Variables:"
> +		rlLog "KERNEL = $KERNEL"
> +		rlLog "KERNEL_PKG_NAME = $KERNEL_PKG_NAME"
> +		rlLog "KERNEL_DEBUGINFO_PKG_NAME = $KERNEL_DEBUGINFO_PKG_NAME"
> +		rpmquery $KERNEL_DEBUGINFO_PKG_NAME
> +		if [ $? -ne 0 ]; then
> +			# we need to install debuginfo for the proper kernel
> +			# but sometimes, debuginfo-install is not available!
> +			which debuginfo-install || rlRun "yum -y install yum-utils dnf-utils" 0 "Installing {yum,dnf}-utils (it has not been present)"
> 
> Could you please detect whether dnf or yum is available and use that instead of hardcoding yum? You can use similar code as the other tests are, e.g.
> https://github.com/CKI-project/tests-beaker/blob/master/networking/common/install.sh#L8
> 
Well, I have thought that this is why DNF has the symlinks:

# ls -l /usr/bin/yum
lrwxrwxrwx 1 root root 5 Nov 19  2018 /usr/bin/yum -> dnf-3
# ls -l /usr/bin/dnf 
lrwxrwxrwx 1 root root 5 Nov 19  2018 /usr/bin/dnf -> dnf-3

> ______________________________________________________________________________________________________________________________________________________________________________________________
> 
> In packages/perf/perftool-testsuite/runtest.sh:
> 
> > +		# log whether we use blacklisting
> +		if [ $PERFTESTS_ENABLE_BLACKLIST -ne 0 ]; then
> +			rlLog "BLACKLISTING ENABLED (known fails will be hidden)"
> +		else
> +			rlLog "BLACKLISTING DISABLED"
> +		fi
> +
> +		# set kptr_restrict to 0
> +		PREVIOUS_KPTR_RESTRICT=`cat /proc/sys/kernel/kptr_restrict`
> +		rlRun "echo 0 > /proc/sys/kernel/kptr_restrict"
> +		rlAssert0 "kptr_restrict must be set to 0" `cat /proc/sys/kernel/kptr_restrict`
> +
> +		# clone the upstream perftool-testsuite repo
> +		fetch_the_testsuite
> +		# return SKip if the testsuite could not be fetched
> +		test -d "perftool-testsuite" || skip_testcase "Could not fetch the upstream testsuite from github. I am sorry, dude."
> 
> Would it make more sense to abort the execution if the upstream testsuite can't be reached? Aborted tests grab attention and not being able to clone the repo might also mean that the repo
> was migrated or removed (extreme, but might happen) and we won't know about it as people don't usually look at the logs of skipped tests.

I am the owner of the repo, so it should not get removed/migrated
without my attention.

The idea is to be able to ignore network problems or github outages.
It's always question whether "no test" means fail or pass, but I
tend to think that CI should be stable and not bother us with false
positives too much. What do you think?

veruu:
> Well, I have thought that this is why DNF has the symlinks: # ls -l /usr/bin/yum lrwxrwxrwx 1 root root 5 Nov 19 2018 /usr/bin/yum -> dnf-3 # ls -l /usr/bin/dnf lrwxrwxrwx 1 root root 5 Nov 19 2018 /usr/bin/dnf -> dnf-3

This is not the case for current rawhide (FC31) composes. We are trying to avoid having to go back and update each test in the future, especially as this is an expected change.

> 
> I am the owner of the repo, so it should not get removed/migrated without my attention. The idea is to be able to ignore network problems or github outages. It's always question whether "no test" means fail or pass, but I tend to think that CI should be stable and not bother us with false positives too much. What do you think?

Our framework treats WARN/ABORTED jobs as infrastructure issues so the developers won't be told "your patch/kernel is bad because something failed to install" (test maintainers shouldn't be getting emails about these tasks either). OTOH the team sees the failed pipelines and can take a look on what caused that and check if there's anything to act upon (e.g. complain about the unstable infrastructure in the lab). The skipped tests are not listed in the report so if people are expecting this test to run they might get confused why it's not the case.

PR #252, opened 2019-07-16
Add test_bpf

sztsian:

### DIFF
--- a/kernel/bpf/test_bpf/runtest.sh
+++ b/kernel/bpf/test_bpf/runtest.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+# Include rhts environment
+. /usr/bin/rhts-environment.sh
+. /usr/share/beakerlib/beakerlib.sh
+. ../../include/install.sh
+
+# Logic:
+# 1. Download the SRPM and install the SRPM
+# 2. Find the source tarball and decompress
+# 3. Extract the lib/test_bpf.c
+# 4. obj-m+= test_bpf.o in Makefile
+# 5. make -C /lib/modules/`uname -r`/build M=`pwd` modules
+# 6. insmod test_bpf.ko
+# 7. journalctl -k --full | sed "s/$(hostname)//g"
+# Expected result
+# Apr 25 04:30:10  kernel: test_bpf: Summary: 349 PASSED, 0 FAILED, [0/341 JIT'ed]
+
+
+DistName=`rpm -E %{?dist} | sed 's/[.0-9]//g'`
+DistVer=`rpm -E %{?dist} | sed 's/[^0-9]//g'`
+DUPARCH=`arch`
+KVer=`uname -r | awk -F '-' '{print $1}'`
+KDIST=`uname -r | sed "s/.$(arch)//g" | awk -F '.' '{print "."$NF}'`
+KBUILD=`uname -r | awk -F '-' '{print $2}' | sed "s/.$(arch)//g" | sed "s/${KDIST}//g"`
+KBuild=${KBuild:-${KBUILD}}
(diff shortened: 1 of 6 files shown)


kernel/bpf/test_bpf/runtest.sh
```diff
@@ -0,0 +1,70 @@
...
+KBuild=${KBuild:-${KBUILD}}
+KBuildPrefix=`echo ${KBuild} | awk -F '.' '{print $1}'`
+
+kernel_nvr=`uname -r | sed "s/.$(arch)//g"`
+KVARIANT=`rpm -q --queryformat '%{sourcerpm}\n' -qf /boot/config-$(uname -r) | sed "s/.srpm//g;s/.src.rpm//g;s/-${kernel_nvr}//g"`
+
+rlJournalStart
+    rlPhaseStartSetup
+        for i in libgcc glibc-static gcc gcc-c++ kernel-devel elfutils-libelf-devel binutils-devel libcap-devel ; do
+            yum install -y $i
```
veruu: Could you please detect whether dnf or yum is available and use that instead of hardcoding yum? You can use similar code as the other tests are, e.g. https://github.com/CKI-project/tests-beaker/blob/master/networking/common/install.sh#L8

kernel/bpf/test_bpf/runtest.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+# Include rhts environment
+. /usr/bin/rhts-environment.sh
+. /usr/share/beakerlib/beakerlib.sh
+
+# Logic:
+# 1. Download the SRPM and install the SRPM
```
veruu: Since this test expects existing SRPM, is it correct to assume that it's only applicable for Brew/koji and RPM builds? For upstream kernels (e.g stable and mainline) we only build tarballs, no RPMs.

sztsian: That also makes sense, but can you point me a way or library function to download the tarball? What I need is only the specific module source code so as long as I have the tarball I can change it.

kernel/bpf/test_bpf/runtest.sh
sztsian: Sure. I'll do it during working hours and rebase the PR.

kernel/bpf/test_bpf/runtest.sh
veruu: You can find some builds at https://artifacts.cki-project.org/builds/
They are installed using the kpkginstall task, specifically this function: https://github.com/CKI-project/tests-beaker/blob/master/distribution/kpkginstall/runtest.sh#L63

sztsian: Great! I'll take a look for this.

sztsian: Hi, I think this differs from what I need. I checked the builds and did not see any source included in the tarballs. Moreover the test_bpf is not pre-build. Given I need the source code, is there a library to do so?

veruu: I don't think there's any tooling for that. We'd have to pass the repo URL to the task, including the URLs of the patches that are being tested and you'd have to apply them in here to reconstruct the "full" tree of tested kernel. It's possible but I don't think it's worth it and the test shouldn't be dealing with things like this, TBH. Thanks for looking into the option!

sztsian: Thanks for the explain.
So I think right now this case will only run against kernel that is delivered by RPM package.

sztsian:
I just did a rebase  to fix the installation part.
For now I think this can only run against rpm-based kernel since I don't have a way to get source code for the other way.

PR #253, opened 2019-07-17
My perf whitelist update 1.2

frozenstein:
aarch64 needs to be added to the "Parse sched tracepoints fields" testcase's blacklist.


veruu:
Applied directly since github's rebase feature is a pain, closing

PR #260, opened 2019-07-19
distribution/kernel-debuginfo: Add regression test

ckalina-rh:
Make sure that kernel-debuginfo vmlinux's .debug* sections are not
accidentally stripped during compile time, as this adversely affects
tracing tools.

### DIFF
--- a/distribution/kernel-debuginfo/runtest.sh
+++ b/distribution/kernel-debuginfo/runtest.sh
@@ -1,0 +1,156 @@
+#!/bin/bash
+set -x
+
+. /usr/bin/rhts_environment.sh
+
+YUM=$(command -v yum)
+if [ -z "$YUM" ]
+then
+	YUM=$(command -v dnf)
+fi
+
+function test_pass()
+{
+	report_result $TEST PASS 0
+}
+
+function test_fail()
+{
+	report_result $TEST FAIL 1
+}
+
+function test_skip()
+{
+	report_result $TEST SKIP 0
+}
+
+function nvr_kernel_current()
+{
+	rpm -qa "kernel*" | grep "$(uname -r)" | sort | head -n1
+}
+
+function nvr_kernel_latest()
+{
+	$YUM list --available kernel{,-rt}.$(uname -m) --showduplicates \
+	| sed "s/.$(uname -m)//" \
+	| sort -k1,1 -k2V \
+	| awk 'END { print $1"-"$2".'$(uname -m)'"; }'
+}
+
+# Determine which kernel debuginfo to download.
+# By default, use currently running kernel (unless overriden from env).
+# This can be achieved either by specifying KERNEL_NVR directly, or one
+# of the following two test config options:
+# CONFIG_NVR_KERNEL_{CURRENT,LATEST}
+
+if [ -z "${KERNEL_NVR++}" ]
+then
+	echo "DEBUG: Using current kernel nvr."
+	KERNEL_NVR="$(nvr_kernel_current)"
+else
+	echo "DEBUG: Using user provided kernel NVR: $KERNEL_NVR"
+fi
+
+if [ -n "$CONFIG_NVR_KERNEL_CURRENT" ]
+then
+	echo "DEBUG: Using current kernel nvr."
+	KERNEL_NVR="$(nvr_kernel_current)"
+fi
+
+if [ -n "$CONFIG_NVR_KERNEL_LATEST" ]
+then
+	echo "DEBUG: Using latest available kernel."
+	KERNEL_NVR="$(nvr_kernel_latest)"
+fi
+
+if [ -z "${KERNEL_NVR++}" ]
+then
+	echo "ERROR: Unable to determine kernel nvr."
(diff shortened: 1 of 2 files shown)


distribution/kernel-debuginfo/runtest.sh
```diff
@@ -0,0 +1,145 @@
...
+	     $CONFIG_NVR_KERNEL_CURRENT
+	echo "DEBUG: CONFIG_NVR_KERNEL_LATEST: " \
+	     $CONFIG_NVR_KERNEL_LATEST
+	echo "DEBUG: rpm -qa \"kernel*\""
+	rpm -qa "kernel*"
+	echo "DEBUG: uname -r: $(uname -r)"
+	echo "DEBUG: yum list --available kernel{,rt}.$(uname -m)" \
+	     "--showduplicates:"
+	yum list --available kernel{,-rt}.$(uname -m) --showduplicates
+	test_fail "FAILED. SEE ERROR OUTPUT ABOVE."
```
veruu: Should this rather abort or skip? Similar with some other cases (e.g. with inability to download debuginfo RPM), the lack of the package would signify a problem in the pipeline rather then a kernel bug that should be reported to patch authors.

distribution/kernel-debuginfo/runtest.sh
```diff
@@ -0,0 +1,145 @@
...
+	echo "DEBUG: CONFIG_NVR_KERNEL_CURRENT: " \
+	     $CONFIG_NVR_KERNEL_CURRENT
+	echo "DEBUG: CONFIG_NVR_KERNEL_LATEST: " \
+	     $CONFIG_NVR_KERNEL_LATEST
+	echo "DEBUG: rpm -qa \"kernel*\""
+	rpm -qa "kernel*"
+	echo "DEBUG: uname -r: $(uname -r)"
+	echo "DEBUG: yum list --available kernel{,rt}.$(uname -m)" \
+	     "--showduplicates:"
+	yum list --available kernel{,-rt}.$(uname -m) --showduplicates
```
veruu: I'm guessing this test should be only ran for RHEL pipelines and not Fedora? We do test RPM builds from Koji but the moment we'd update to fc31 yum will no longer be symlinked so this would break, unless you can make it work for dnf too (and picking the command based on which one is available).

ckalina-rh: In principle, the test makes sense for both. Regarding yum, I can just use something like:
```
YUM=$(command -v yum)
[[ -z $YUM ]] && YUM=$(command -v dnf)
```
and use that variable instead.

veruu: Having it for both would be great!

distribution/kernel-debuginfo/runtest.sh
ckalina-rh: Sure. Looking at valid results for rhts-report-result, "Valid values are PASS, WARN, FAIL, or SKIP."; so I can just use
```
function test_skip() { report_result $TEST SKIP 0; }
# ...
test_skip ...
```
Or would you prefer WARN?

veruu: We use WARN/ABORTED to mark infrastructure issues -- e.g. the test can't be compiled, the network is down in the lab.

We use SKIP/COMPLETED to mark the test requirements are not met (e.g. a debug package is required for the test but we don't build it).

So in this case, SKIP seems to be making more sense.

distribution/kernel-debuginfo/runtest.sh
ckalina-rh: Changes present in v2.

distribution/kernel-debuginfo/runtest.sh
ckalina-rh: Changes present in v2.

ckalina-rh:
Force-pushed v2 version of the patch. Above remarks were addressed.

distribution/kernel-debuginfo/runtest.sh
```diff
@@ -0,0 +1,156 @@
...
+	echo "DEBUG: CONFIG_NVR_KERNEL_LATEST: " \
+	     $CONFIG_NVR_KERNEL_LATEST
+	echo "DEBUG: rpm -qa \"kernel*\""
+	rpm -qa "kernel*"
+	echo "DEBUG: uname -r: $(uname -r)"
+	echo "DEBUG: $YUM list --available kernel{,rt}.$(uname -m)" \
+	     "--showduplicates:"
+	$YUM list --available kernel{,-rt}.$(uname -m) --showduplicates
+	test_skip "FAILED. SEE ERROR OUTPUT ABOVE."
+	exit 1
```
veruu: Restraint interprets non-zero exit codes as a failure and thus this overrides the SKIP. For skip to propagate correctly the exit codes needs to be 0.

ckalina-rh: Changed in v3.

ckalina-rh:
Force pushed v3 to address the return code override issue.

PR #261, opened 2019-07-22
Port test case 'power-management/rapl/powercap' (#1017)

idorax:
This is the 1st PR of ticket `FASTMOVING-1017`, the source code saved in `tests/kernel` includes 7 small shell scripts, I converted them to functions being saved into `runtest.sh` to make things simple and achieve better readability.

A manual test on RHEL8/x86_64 is done, here is the test log:
- http://pastebin.test.redhat.com/781779

### DIFF
--- a/power-management/rapl/powercap/runtest.sh
+++ b/power-management/rapl/powercap/runtest.sh
@@ -1,0 +1,230 @@
+#!/bin/bash
+#
+# Copyright (c) 2019 Red Hat, Inc. All rights reserved.
+#
+# This copyrighted material is made available to anyone wishing
+# to use, modify, copy, or redistribute it subject to the terms
+# and conditions of the GNU General Public License version 2.
+#
+# This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public
+# License along with this program; if not, write to the Free
+# Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+# Boston, MA 02110-1301, USA.
+#
+
+FILE=$(readlink -f ${BASH_SOURCE})
+NAME=$(basename $FILE)
+CDIR=$(dirname $FILE)
+TEST=${TEST:-"$0"}
+TMPDIR=/var/tmp/$(date +"%Y%m%d%H%M%S")
+
+source ${CDIR%/power-management/rapl/powercap}/cpu/common/libbkrm.sh
+
+# XXX: This function is from power-management/rapl/powercap/cap.sh
+function set_cap
+{
+    typeset energy=${1?"*** energy_in_microjoules***"}
+    typeset cpu_sockets=$(lscpu | grep 'Socket(s):' | \
+                          sed 's/^Socket(s):[^0-9]*\([0-9]*\)/\1/')
+    typeset cap_energy=$(( energy / cpu_sockets ))
+    typeset rapl_path="/sys/devices/virtual/powercap/intel-rapl"
+    typeset rapl=$(ls -1 $rapl_path | grep 'intel-rapl:')
+    typeset cpu=""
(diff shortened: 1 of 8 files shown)


power-management/rapl/powercap/runtest.sh
```diff
@@ -0,0 +1,201 @@
...
+rlJournalStart
+    rlPhaseStartSetup
+        rlRun "TmpDir=\$(mktemp -d)" 0 "Creating tmp directory"
+    rlPhaseEnd
+
+    rlPhaseStartTest
+        #debug
+        rlRun -l "find /sys/devices/ -name *rapl*"
+        rlRun -l "lsmod"
+        rlRun -l "modprobe intel-rapl"
```
veruu: Does this test apply only to Intel CPUs? If so, should we check the vendor and SKIP if it's not applicable instead of failing the test when Intel-specific module fails to load?

power-management/rapl/powercap/runtest.sh
```diff
@@ -0,0 +1,201 @@
...
+    set_cap $old_cap
+
+    if (( e_full_cap > measured_max || e_full_cap < measured_min )); then
+        rlLog "FAIL: measured energy consumption doesn't match capped value"
+        return 1
+    else
+        rlLog "PASS: Test passed"
+    fi
+
+    # All my tests can write warnings (like non met depedencies) to this
```
veruu: If the dependencies aren't met, can we instead ABORT or SKIP based on what it is? We don't want to tell developers their patches are wrong because a test dependency wasn't met.

power-management/rapl/powercap/runtest.sh
```diff
@@ -0,0 +1,201 @@
...
+    # file, which wil be translated to warning in the Beaker.
+    touch warn.tmp
+}
+
+# === power-management/rapl/powercap/warn.sh ==================================
+function powercap_warn
+{
+    typeset warn_file="./warn.tmp"
```
veruu: Grepping the test for `warn_file` and `warn.tmp`, I can only find the file creation and then this processing while there is no command that would actually modify the file. Is this done by default somehow? If not, can we just remove the function please?

idorax: Removed as it is useless.

power-management/rapl/powercap/runtest.sh
idorax: Have added a function `is_intel()` to startup(). If the CPU is not intel, will abort the case.

idorax:
Please find the latest test log for RHEL8/x86_64 [here](http://pastebin.test.redhat.com/782109).

power-management/rapl/powercap/runtest.sh
```diff
@@ -0,0 +1,230 @@
...
+    fi
+
+    is_intel
+    if (( $? != 0 )); then
+        rlSetReason $BKRM_UNSUPPORTED "non-intel CPU is unsupported"
+        return $BKRM_UNSUPPORTED
+    fi
+
+    if [[ ! -d $TMPDIR ]]; then
+        rlRun "mkdir -p -m 0755 $TMPDIR" || return $BKRM_FAIL
```
veruu: Shouln't this be abort rather then fail? Not that it would be expected to fail but it's a preparation for test, rather then part of the test itself.

veruu: Other than this, the changes look good!

idorax: Okay, use `return $BKRM_FATAL` instead of `return $BKRM_FAIL`. BTW, 
* Also `return $BKRM_FATAL` when executing `rlRun -l "modprobe intel-rapl"` because it is also a preparation for test;
* File `cpu/driver/runtest.sh` and `cpu/idle/runtest.sh` are updated too as they have the same code line.

idorax: Just now I renamed `BKRM_FATAL` to be `BKRM_UNINITIATED` as word `UNINITIATED` is more precise to describe the test code that is in preparation phase before a real test is started. For details, please refer to:
* http://pastebin.test.redhat.com/782545

PR #263, opened 2019-07-23
disable ltp syslog01 test since rsyslog is disabled by default on F30

rasibley:


distribution/ltp/lite/configs/RHELKT1LITE.20190115
```diff
@@ -1233,7 +1233,8 @@ sysfs04 sysfs04
 sysinfo01 sysinfo01
 sysinfo02 sysinfo02
 
-syslog01 syslog01
+# rsyslog is disabled by default on F30
+#syslog01 syslog01
```
veruu: Would it be better to check if we're running on Fedora 30 in the knownissue script and exclude the test that way, so we can still run it on the other releases?

rasibley:
@veruu good point, we do run this version for ark kernels as well using rhel8, I adjusted it, however I probably should test it to make sure I didn't break it :-)

veruu:
This should be a distro thing rather then kernel if I'm not mistaken (and new Fedora should be using systemd's journaling). If the ARK testing will continue to use RHEL8 then we should be ok, I don't remember the test failing there.

veruu:
@rasibley LGTM, let me know if you want to merge it now or you want to do some testing first

rasibley:
@veruu ok it's ready now, just ran a few test jobs and it passed, I had to disable all syslog tests since the others were failing too.

PR #265, opened 2019-07-24
Add ftrace-tracer

sztsian:
Ported from
https://github.com/sztsian/kernel-sanity/tree/master/ftrace-tracer
This test enables each ftrace tracer and then disable it.
Tracers should be able to be enabled, and no hang or panic should happen.


rasibley:
Thanks @sztsian this was a lot of code to review, I am not seeing any obvious reason we can't merge, from what I can understand :-) I don't see any infra failures which would cause a kernel to fail where we may want to skip/abort instead or any confidential information. @veruu did you have time to review ?

PR #266, opened 2019-07-25
sync serge-testsuite with downstream

rasibley:


packages/selinux-policy/serge-testsuite/runtest.sh
```diff
@@ -104,9 +112,31 @@ rlJournalStart
...
+                rlRun "git config --global user.email nobody@redhat.com"
+                rlRun "git config --global user.name 'Nemo Nobody'"
+                rlRun "git checkout -b testing-cherry-picks" 0
+                break
+            done
+            for pull in $GIT_PULLS; do
+                ref="refs/pull/$pull/head"
+                if ! rlRun "git fetch origin $ref:$ref" 0; then
+                    rlRun "git checkout $GIT_BRANCH" 0
+                    rlWarn "PR merge failed, falling back to GIT_BRANCH"
```
veruu: Are we sure we want to report warn here and below since it will mark the job as failing and send a failure report to developers? I'm thinking it might be better to just fall back (and log the message without warning), what do you think?

rasibley:
Thanks @veruu good catch, I updated it to use rlLog function.

PR #267, opened 2019-07-25
initial porting suspend resume

ftjahjad:


power-management/suspend-resume/PURPOSE
```diff
@@ -0,0 +1,2 @@
+Try to suspend and resume the system.
+This test requires non hypervisor machine.
```
veruu: I don't see any checks for this in the test itself. Shouldn't a simple check and skip be present?

Other than this, LGTM

rasibley: hi @veruu just an fyi this will be running in a recipe targeting bare metal machines, however doesn't hurt to have that adittional check. Otherwise looks good to me!

ftjahjad: Hi @veruu , and @rasibley,

Added checking, if not baremetal machine, the test will skip. 
please review the changes.

PR #286, opened 2019-08-02
Fix invalid raid0 module added in RHEL7.x (x >= 4)

zhongchanghui:


storage/swraid/trim/main.sh
```diff
@@ -28,8 +28,8 @@ function runtest()
 {
 	rlRun "modprobe raid456 devices_handle_discard_safely=Y"
 	rlRun "echo Y >/sys/module/raid456/parameters/devices_handle_discard_safely"
-	uname -r | grep el7
-	if [ $? -eq 0 ];then
+	typeset release=$(cat /etc/redhat-release | tr -cd "[0-9.]")
+	if [[ $(bc -l "$release <= 7.4") -eq 1 ]]; then # release <= 7.4 is true
```
veruu: Hi,
I've tried the copied commands on RHEL7.6 and I'm getting the following output:

```
# release=$(cat /etc/redhat-release | tr -cd "[0-9.]")
# if [[ $(bc -l "$release <= 7.4") -eq 1 ]]; then echo "condition passed" ; fi
File 7.6 <= 7.4 is unavailable.
```
I can reproduce the behavior on FC29 as well.
Man page of `bc -l` says `Define the standard math library.` so I think this is not the option you wanted to use?

zhongchanghui: Hello,Veronika
Sorry, some of the code is missing in the submitted PR. The correct one should be "bc -l <<< "$release <= 7.4"". I have submitted a new commit to fix this error. I am verified at RHEL7.4, 7.7, 8.1 and F30.
Thanks

PR #288, opened 2019-08-02
Make sure beaker task is not marked as 'Aborted' if test fails

idorax:
Currently `main()` in `libbkrm.sh` returns $rc which will cause beaker task is marked as 'Aborted' if `runtest()` fails. Hence, `main()` is updated to always return 0, and two status code (`BKRM_STATUS_COMPLETED`  and `BKRM_STATUS_ABORTED`) are defined. BTW, `rhts-abort -t recipe` is removed from rlAbort() after reading the [manual page of rhts-abort](https://beaker-project.org/docs/user-guide/task-environment.html#commands) , please correct me if my understanding is not right.

### DIFF
--- a/cpu/common/libbkrm.sh
+++ b/cpu/common/libbkrm.sh
@@ -31,6 +31,12 @@ BKRM_UNSUPPORTED=2 # should go to rlSkip()
 BKRM_UNINITIATED=3 # should go to rlAbort()
 
 #
+# Status code definitions
+#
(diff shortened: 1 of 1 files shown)


veruu:
@cki-bot test please!

idorax:
Failures reported by cki-bot are not related to this PR. There are three beaker jobs for unit testing, and the test results are expected.
* Test 'FAIL': `3708729`
* Test 'Abort': `3708725`
* Test 'Skip':  `3708722`

PR #293, opened 2019-08-05
Rename kernel dir to trace

### LINKED ISSUE #287: Rename kernel/bpf kernel/ftrace kernel/lib since this only contains kernel tests
As per discussion with Vector Li, this repo only contains kernel tests, and he suggest me to rename my ftrace and bpf tests to avoid the useless folder structure kernel. I'll do this with a PR after I tested it.

sztsian:
Fixes #287


veruu:
This PR requires changes to kpet-db as the test fetch URLs will be different. It would be best if both changes could be merged at the same time to avoid errors. I'm not on VPN right now so I can't merge both right now (or even check if there is a PR for kpet-db opened). cc @rasibley

rasibley:
I filed a kpet MR to change the path, will merge these at the same time. The ftrace one will be onboarded, it still has a pending MR.

PR #294, opened 2019-08-06
kdump: install and config kexec-tools for Fedora

emmawu:
In Fedora, kdump is not installed and enabled by default. 
So added steps to 
1) Install kexec-tools and enabled kdump service.  
2) Configure default crashkernel memory

Test passed on Fedora30.


rasibley:
hi @emmawu does this mean kdump test is now supported in Fedora and upstream kernels, currently it's disabled but please let me know if I can re-enable in kpet-db for a given arch.

emmawu:
Hi @rasibley,
Hmm, it only means the kdump test is now runnable on Fedora.
We disable kdump tests on Fedora and upstream kernel because
1) RHEL8 kexec-tools would fail on upstream kernel on ppc64le/x86_64. This is still an known issue..
2) For test ran on Fedora as base distro, it's actually testing the Fedora kexec-tools and Fedora kernel (or upstream kernel, depends on how tests are triggered). QE is currently out of capacity of tracking Fedora kexec-tools failures.

PR #298, opened 2019-08-08
blktests: remove the blacklisted nvme/ test case

yizhanglinux:
upstream blktests have some update which may fix the test case issue, so
re-add the test case

Signed-off-by: Yi Zhang <yi.zhang@redhat.com>


rasibley:
Thanks for re-enabling the tests !

PR #311, opened 2019-08-16
IOMMU boot test

emanchado:
Add a IOMMU boot test with different parameters. It works on RHEL8, RHEL7, and Fedora 30 at least. It will have to be adapted to work on older RHEL releases due to the use of `journalctl`, though.

### DIFF
--- a/iommu/boot/runtest.sh
+++ b/iommu/boot/runtest.sh
@@ -1,0 +1,164 @@
+#!/bin/bash
+# vim: dict=/usr/share/rhts-library/dictionary.vim cpt=.,w,b,u,t,i,k
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   runtest.sh of /kernel/iommu/boot
+#   Description: Test various IOMMU boot options
+#   Author: William Gomeringer <wgomerin@redhat.com>
+#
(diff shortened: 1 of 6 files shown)


emanchado:
@cki-bot test this stuff, please!

iommu/boot/runtest.sh
```diff
@@ -0,0 +1,164 @@
...
+    dmesgErrors
+else
+    if [[ $cpuvendor = "GenuineIntel" ]]; then
+	bootOptions $DefaultBootOptionsIntel
+	dmesgErrors
+    elif [[ $cpuvendor = "AuthenticAMD" ]]; then
+       	bootOptions $DefaultBootOptionsAMD
+	dmesgErrors
+    else
+	report_result "${TEST}/nonAMDorIntelProcessor" "FAIL" 0
```
veruu: Should this be SKIP instead, as it's not the patch developer's fault that the test is running on unsupported machine?

veruu:
@emanchado just FYI, if the test wasn't already added to kpet-db, the generated XML won't contain this test and it won't run. /me is not on VPN to check that

iommu/boot/runtest.sh
emanchado: Sounds good. That was in the original test, and I didn't even know `SKIP` was a thing in this harness :sweat_smile: 

Which reminds me: how much are we supposed to review/change the tests when copying them?

PR #312, opened 2019-08-16
FMK-1034 : Enable /kernel/networking/vnic/ipvlan/basic test in kpet-db

idorax:


networking/vnic/ipvlan/basic/runtest.sh
```diff
@@ -0,0 +1,1261 @@
...
+
+if ip link add link $TEST_IFACE name ipvlan1 type ipvlan mode l3
+then
+	ip link del ipvlan1
+	for test_item in $TEST_ITEMS
+	do
+		$test_item
+	done
+else
+	echo "not support"
```
veruu: Should we report a skip here?

idorax: Yes, added code to skip it.

rasibley:
@veruu is on PTO so I can finish the review here, I see you added the skip option per her suggestion, looking over the rest of the code I don't see any other reason not to merge, it can be tested more thoroughly in kpet-db.

PR #316, opened 2019-08-20
FMK-1034 : Enable /kernel/networking/vnic/ipvlan/basic test in kpet-db (cont)

idorax:
Add necessary C files used by multicast test
1. networking/common/src/mtools/mdump.c
1. networking/common/src/mtools/msend.c
1. networking/common/src/mtools/mpong.c


idorax:
A beaker job is done, and the job id is `3738486`, test result of multicast looks good.

PR #318, opened 2019-08-22
FMK-1035 : Enable /kernel/networking/bridge/sanity_check test in kpet-db

idorax:


idorax:
Test results on x86 look good.
-    RHEL8: `3743196`
-    RHEL7: `3743204`
-    Fedora30: `3743202`

PR #322, opened 2019-08-23
cpu/idle: avoid running test on systems which do not support mwait

darcari:
Any system that does not support the mwait instruction will not have an
idle driver running.  As such, idle results will be poor in comparison
to systems with an idle driver.  Avoiding executing the test on these
systems.

Also, cleanup SKIP cases to match coding standard.


darcari:
Decided to code somewhat differently.

PR #324, opened 2019-08-26
networking: clean up dummy interface after bridge test

linup2011:
remaining dummy interface would cause driver-sanity fail, so clean up dummy used in bridge test


idorax:
The fix looks good to me. And it is tested via job `3749753`.

PR #335, opened 2019-08-28
Debug kernel install support for kpkginstall

major:


distribution/kpkginstall/runtest.sh
```diff
@@ -38,14 +44,29 @@ function set_package_name()
       PACKAGE_NAME=kernel
   fi
 
+  # Append "-debug" if we were asked to install the debug kernel.
+  if [[ "${KPKG_INSTALL_DEBUG:-}" == "yes" ]]; then
```
emanchado: Hm, why the `:-`? if you're not providing any default, isn't it the same to write the following?
```
if [[ "$KPKG_INSTALL_DEBUG" == "yes" ]]
```

distribution/kpkginstall/runtest.sh
```diff
@@ -38,14 +44,29 @@ function set_package_name()
...
+    echo "Debug kernel was requested -- appending -debug to package name"
+    PACKAGE_NAME=${PACKAGE_NAME}-debug
+  fi
+
+  # Write the PACKAGE_NAME to a file in /tmp so we have it after reboot.
+  echo -n "${PACKAGE_NAME}" | tee -a /tmp/kpkginstall/KPKG_PACKAGE_NAME
```
emanchado: I wonder if `/tmp/kpkginstall/KPKG_PACKAGE_NAME` and co. should be stored in "constants"...

distribution/kpkginstall/runtest.sh
```diff
@@ -38,14 +44,29 @@ function set_package_name()
...
+  # Recover the saved package name from /tmp/KPKG_KVER if it exists.
+  if [ -f "/tmp/kpkginstall/KPKG_KVER" ]; then
+    KVER=$(cat /tmp/kpkginstall/KPKG_KVER)
+    return
+  fi
+
   if [[ "${KPKG_URL}" =~ .*\.tar\.gz ]] ; then
     declare -r kpkg=${KPKG_URL##*/}
-    tar tf "$kpkg" | sed -ne '/^boot\/vmlinu[xz]-[1-9]/ {s/^[^-]*-//p;q}; $Q1'
+    KVER=$(tar tf "$kpkg" | sed -ne '/^boot\/vmlinu[xz]-[1-9]/ {s/^[^-]*-//p;q}; $Q1')
```
emanchado: What is `$Q1`? Is that a sed thing? I don't remember having seen it before.

distribution/kpkginstall/runtest.sh
```diff
@@ -320,22 +372,44 @@ else
...
     rhts-abort -t recipe
     exit 0
   fi
+
+  echo "✅ Found the correct kernel version running!"
+
+  # We have the right kernel. Do we have any call traces?
+  dmesg | grep -qi 'Call Trace:'
+  dmesgret=$?
+  if [[ -n "${CHECK_DMESG}" && ${dmesgret} -eq 0 ]]; then
```
emanchado: Why `${dmesgret}` instead of simply `$dmesgret`? Is it a style thing? Because if so it doesn't seem consistently applied :frowning_face:

distribution/kpkginstall/runtest.sh
```diff
@@ -320,22 +372,44 @@ else
...
+    DMESGLOG=/tmp/dmesg.log
+    dmesg > ${DMESGLOG}
+    rhts_submit_log -l ${DMESGLOG}
```
emanchado: As we're assuming bash anyway, we can AFAIK simplify these two lines and write:
```
rhts_submit_log -l <(dmesg)
```

veruu: Not sure if this will work as `rhts_submit_log` signature expects a filename and then uploads that file under the same name into Beaker result logs :thinking:

distribution/kpkginstall/runtest.sh
veruu: Not me -- Artem (the original author) used it and I kept it there with a "it works, don't touch it" logic. However, doing some research it looks like it means "if we got this far quit with exit code 1". So the code tries a few matches to get the kernel version and if none of them match, returns an unsuccessful return code.

The broken down logic looks like `$` -- got to last line; `Q` -- command to quit; `1` -- parameter to quit command

major:
See internal job 3757023 for successful debug kernel installs.

major:
See internal job 3757188 for a full test across all kernels.

PR #336, opened 2019-08-29
trace/bpf: Improve support for debug kernel

sztsian:


sztsian:
Hold on for a while, let me add some more support via this PR.

sztsian:
Ok, also added the workaround in case kernel-devel is cross-compiled.

PR #342, opened 2019-08-30
WIP: move cpu/common/libbkrm.sh to cki_lib/libbkrm.sh

idorax:
We have **cki_lib** now, I'd like to move file `cpu/common/libbkrm.sh` to directory `cki_lib` as this file is used by test cases of both **cpu** and **power-management**.


idorax:
Please hold on review, I'm debugging the changes as dir **cki_lib** is empty when running test cases via beaker.

darcari:
cki_lib was populated for my test:

# pwd
mnt/tests/github.com/darcari/tests-beaker/archive/die.v3.zip
# ls
cki_lib  cpu
# ls cki_lib/
lib.sh

I added:
	@echo "RepoRequires: cki_lib" >> $(METADATA)

To the cpu/die Makefile.

rasibley:
@darcari  @idorax  What I'd like to see is abort_task and skip_task added to this file here: https://github.com/CKI-project/tests-beaker/blob/master/cki_lib/lib.sh using a similar coding style as abort_function. There are other functions included in https://github.com/CKI-project/tests-beaker/blob/master/cpu/common/libbkrm.sh which are wrapping other functions which seem a bit complex, and I would rather keep it simple if possible, how do others feel ?

darcari:
You'd need to modify the libbkrm.sh code to use the lib.sh code.  Although abstract, the main function nicely masks a lot of the beakerness of these tests.  As this may appeal to other developers it seems like it belongs in cki_lib as well, but I agree that putting this in  a different file seems reasonable.  Coding style should match across lib_cki IMHO.

rasibley:
hi @idorax can you please move everything but main() to cki_lib/lib.sh ? Also there will be some duplicating and different coding styles, abort_recipe vs rlAbort, can we instead use existing abort_recipe() and please change rlSkip to skip_recipe() using the same coding style as abort_recipe() ? Also please include a beaker ID for a test job which will includes all related tests which are conforming to this change, e.g. kpginstall, cpu and powermgmt cases. Also please remove WIP state after if you feel it's ready so we can merge :-)

idorax:
Hi @emanchado, I'm debugging my code because hitting a very weird problem that cki_lib/lib.sh is missing when running a test case. e.g.
- http://pastebin.test.redhat.com/797295

So, please help to review once 'WIP:' is removed.

idorax:
- http://pastebin.test.redhat.com/797295

PR #344, opened 2019-08-30
Remove references to OUTPUTFILE

major:
We never submit the log referenced by `$OUTPUTFILE`, so we can remove
those references. The main `taskout.log` is enough for kpkginstall.


major:
Submitted internal job 3758537 to test this one and #345.

PR #345, opened 2019-08-30
 Check command success directly in if/else

major:
The `shellcheck` linter stronly recommends checking for command success
and failure directly in the `if/else` statement itself rather than checking
for `$?`. It requires less code and allows us to easily print some
success output with emojis.

⚠️ **Requires https://github.com/CKI-project/tests-beaker/pull/344 to merge first!**


major:
Submitted internal job 3758537 to test this one and #344.

PR #346, opened 2019-08-30
Use abort_recipe() for failures

major:
Reduce repeated code by consolidating all of the failure handling in a
single function that has emojis.

:warning: **Requires #344 and #345 to merge first!**

PR #347, opened 2019-08-30
 Get package name/debug from KPKG_URL variables

major:
Getting package names from the repository packages is problematic. Most
pipeline jobs have a `package_name` variable that can be passed via
`KPKG_URL`.

These variables can be passed in the URL after a pound sign `#` and
parsed by kpkginstall.

We can now handle URLs like these:

*  example.com/job/123/repo/#package_name=kernel-rt&amp;debug=yes

:warning: **Requires #346 to merge first.**


major:
Testing via 3758698.

major:
Testing in 3758732.

PR #350, opened 2019-08-30
Make kpkginstall messages friendly

major:
The kpkginstall output is highly cluttered and confusing to people outside
the CKI group.

Add simple functions for printing messages with different statuses and use
those throughout the script. Also, remove `-x` from the script to reduce
clutter in the output.

### DIFF
--- a/cki_lib/lib.sh
+++ b/cki_lib/lib.sh
@@ -35,3 +35,20 @@ function cki_yum_tool()
   fi
 }
 
+function print_info()
+{
+  # Print an informational message with a friendly emoji.
+  echo "ℹ️ ${1}"
+}
+
+function print_success()
+{
+  # Print a success message with a friendly emoji.
+  echo "✅ ${1}"
+}
+
+function print_warning()
+{
+  # Print an warning message with a friendly emoji.
+  echo "⚠️ ${1}"
+}
--- a/distribution/kpkginstall/runtest.sh
+++ b/distribution/kpkginstall/runtest.sh
@@ -1,6 +1,4 @@
 #!/bin/bash
-set -x
-
 
 . /usr/bin/rhts_environment.sh
 
@@ -40,7 +38,8 @@ function parse_kpkg_url_variables()
   # all keys uppercase for consistency.
   for ((i=0; i<${#parm[@]}; i+=2))
   do
-      readonly KPKG_VAR_${parm[i]^^}=${parm[i+1]}
+    print_success "Found URL parameter: ${parm[i]^^}=${parm[i+1]}"
+    readonly KPKG_VAR_${parm[i]^^}=${parm[i+1]}
   done
 }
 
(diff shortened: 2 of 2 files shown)


major:
Testing in 3758882.

cki_lib/lib.sh
```diff
@@ -35,3 +35,20 @@ function cki_yum_tool()
...
+function print_success()
+{
+  # Print a success message with a friendly emoji.
+  echo "✅ ${1}"
+}
+
+function print_warning()
+{
+  # Print an warning message with a friendly emoji.
+  echo "🟡 ${1}"
```
emanchado: Hm, this emoji doesn't seem to be recognised by either GitHub, or my terminal :cry: I wonder if we should use a different one, for better compatibility. Weirdly, looking further down in the PR, it seems that the emoji used by hand (⚠️) is recognised? So maybe use that one? idk

Also, maybe drop the comments? I don't think they give that much value :sweat_smile:

distribution/kpkginstall/runtest.sh
```diff
@@ -160,34 +162,35 @@ function get_kpkg_ver()
...
   if [ -z "${KVER}" ]; then
     abort_recipe "Failed to extract kernel version from the package" FAIL
   else
-    echo "Kernel version is ${KVER}"
+    print_success "Kernel version is ${KVER}"
   fi
 
   if tar xfh ${kpkg} -C / 2>&1; then
-    echo "✅ Extracted kernel package successfully: ${kpkg}"
+    print_success "✅ Extracted kernel package successfully: ${kpkg}"
```
emanchado: The first emoji should be dropped from here, yeah?

distribution/kpkginstall/runtest.sh
```diff
@@ -259,14 +265,18 @@ function select_yum_tool()
...
     ALL="all"
-    ${YUM} install -y yum-plugin-copr
+    COPR_PLUGIN_PACKAGE=yum-plugin-copr
   else
     abort_recipe "No tool to download kernel from a repo" WARN
   fi
+
+  print_info "Installing package manager prerequisites"
+  ${YUM} install -y ${COPR_PLUGIN_PACKAGE} > /dev/null
+  print_success "Package manager prerequisites installed successfully"
```
emanchado: Nice little refactoring! :hand:

distribution/kpkginstall/runtest.sh
```diff
@@ -304,65 +313,73 @@ function copr_prepare()
...
   fi
+  print_info "Installing kernel firmware package"
+  $YUM install -y $FIRMWARE_PKG > /dev/null
+  print_success "Kernel firmware package installed"
 
   # Workaround for BZ 1698363
   if [[ "${ARCH}" == s390x ]] ; then
-    grubby --set-default /boot/"${KVER}" && zipl
+    grubby --set-default /boot/"${KVER}" > /dev/null
+    zipl > /dev/null
```
emanchado: Hm... was it on purpose to execute `zipl` regardless of the success of `grubby`?

cki_lib/lib.sh
major: I changed to the yellow circle since the warning triangle wasn't showing up well in some browsers. :/ I'll switch it back and try again.

distribution/kpkginstall/runtest.sh
major: Good catch!

distribution/kpkginstall/runtest.sh
major: Oops, I didn't meant to do that.

major:
Testing in job 3763627 internally.

PR #351, opened 2019-09-02
Disable case "masec between 2 netns" because system panicked due to bz1745880

idorax:
Fei Liu helped to provide the patch, please review, thanks!


Feliu-C:
This bug was discovered by macsec test, which caused panic, so we temporarily disable it.

idorax:
Thanks to @Feliu-C, I have added some comments to file `runtest.sh`.

rasibley:
hi @idorax thanks for adding a comment to block out that section, LGTM!

PR #357, opened 2019-09-05
vm/kvm-unit-tests: Run KVM Unit Tests with Advanced Virt (qemu-4.0)

mcondotta:
This patch will enable the KVM Unit Tests suite to run with both the regular QEMU (qemu-2.12) as well as the Advanced Virt QEMU (qemu-4.0).


vm/kvm-unit-tests/runtest.sh
```diff
@@ -165,6 +165,38 @@ cat test.log | tee -a $OUTPUTFILE
 # check for any new failures
 grep FAIL test.log >> failures.txt
 
+cat >/etc/yum.repos.d/rhel8-advvirt.repo <<EOF
```
rasibley: Hi @mcondotta thanks for the PR :-) do you think the test should disable the repo after running the test to properly cleanup since this test is run in conjunction with other tests ?

mcondotta: Hello @rasibley thank you very much for the review.
I've added the cleanup code to remove the Repo and downgrade the QEMU version.

rasibley: Thanks looks good!

vm/kvm-unit-tests/runtest.sh
```diff
@@ -165,6 +165,45 @@ cat test.log | tee -a $OUTPUTFILE
 # check for any new failures
 grep FAIL test.log >> failures.txt
 
+# Add Advanced VIRT repo and update QEMU version
+cat >/etc/yum.repos.d/rhel8-advvirt.repo <<EOF
+[rhel8-advvirt]
+name=rhel8-advvirt
+baseurl=http://download.devel.redhat.com/rhel-8/rel-eng/ADVANCED-VIRT-8/latest-ADVANCED-VIRT-8.1.0-RHEL-8/compose/Advanced-virt/$(uname -i)/os/
```
rasibley: Is there an external repo we can point to ? If not I have another suggestion where we can enable the repo in kpet-db.

rasibley: hi @mcondotta we have the repo enabled by default using kpet now, so you can remove that part about creating the yum repo, and instead enable/disable the repo before/after testing. This is enabled only for rhel8 testing, I see this test is also enabled on ark kernel which runs on fedora, not sure if changes are needed there as well ?

vm/kvm-unit-tests/runtest.sh
```diff
@@ -165,6 +165,38 @@ cat test.log | tee -a $OUTPUTFILE
 # check for any new failures
 grep FAIL test.log >> failures.txt
 
+# Run KVM Unit tests with Advanced Virt (qemu-4.0) if possible
+if dnf repolist --all | grep rhel8-advvirt; then
```
rasibley: The repo is disabled by default, you will need to enable it first, then please disable it after testing.

vm/kvm-unit-tests/runtest.sh
rasibley: hi @mcondotta this has been opened for a while, I never saw a response to my last comment, any further updates ? :-)

rasibley:
hi @mcondotta this has opened since Sept 5th :-( did you plan to update it or should we close ?

rasibley:
@mcondotta is this still needed ?

PR #358, opened 2019-09-05
install the right kernel-devel for running kernel

rasibley:
@major can you review ? I decided to remove that code since it's installed by default as part of kpginstall, otherwise the harness will install it from the Makefile. See beaker ID 3769092 for test results.


major:
This will cause delays during the harness step since some of these packages won't exist and restraint will go through them one by one, but it's no worse than what we have right now. :slightly_smiling_face:

rasibley:
@major well what happens now is restraint just skips it all together since the packages are already installed.

PR #359, opened 2019-09-05
another python2 fix for kernel/sound/aloop

jkysela-rh:


sound/aloop/data.py
```diff
@@ -10,7 +10,7 @@
   for i in range(0, frames):
     s = i & 65535
     frame = [(s >> 8) & 255, s & 255, ((s >> 8) & 255) ^ 255, (s & 255) ^ 255]
-    fp.write(bytes(frame))
+    fp.write(bytearray(frame))
```
rasibley: @jkysela-rh thank you, will this fix it for both python2 and python3 ?

jkysela-rh: Yes, I verified the fix with both python2 and python3.

rasibley: Thanks just checking :-) I'll merge and we can re-trigger the testing in kpet CI.

PR #361, opened 2019-09-05
remove kernel-devel dep, already installed with kpginstall

rasibley:
@major removing them fixes it since the correct devel rpm is already installed. See J:3769878 and J:3769874


major:
LGTM!

PR #362, opened 2019-09-06
tunnel: wait longer to make case more stable

linup2011:


linup2011:
> I know nothing about this test, but I'm wondering: isn't there a way to check if `socat` is ready? Or maybe it would be possible to run the code that tries to send something to that port five times, waiting one second after each, and breaking from the loop when it succeeds?
> 
> Anyway, otherwise LGTM. As I don't know if it's realistic to do the above, I'll just approve.

as I run socat in background, we can check if it is running, but it can't guarantee that everything is ok, and that's why the case may fail, also why I submit this patch

PR #363, opened 2019-09-07
add non bare-metal patch and link known issues

rasibley:
Update to a later commit, link known issues, apply patch for non bare-metal machines as timing on systems with shared resources (and high steal time) is not accurate. See J:3771976 for results, one job failed with an intermittent failure, will continue to monitor it.


rasibley:
@major can you review ?

PR #365, opened 2019-09-09
initial porting pjd-fstest

ftjahjad:
Initial porting for pjd-fstest (ext3/ext4/xfs)

### DIFF
--- a/filesystems/general/pjd-fstest/README.md
+++ b/filesystems/general/pjd-fstest/README.md
@@ -1,0 +1,13 @@
+# pjd-fstest filesystem test suite
+Tests POSIX features of file systems. \
+Test Maintainer: [Zorro Lang](mailto:zlang@redhat.com)
+
+## How to run it
+
+### Dependencies
+Please refer to the top-level README.md for common dependencies. Test-specific dependencies will automatically be installed when executing 'make run'. For a complete detail, see PURPOSE file
+
+### Execute the test
+```bash
+$ make run
+```
--- a/filesystems/general/pjd-fstest/runtest.sh
+++ b/filesystems/general/pjd-fstest/runtest.sh
@@ -1,0 +1,224 @@
+#!/bin/bash
+# vim: dict=/usr/share/rhts-library/dictionary.vim cpt=.,w,b,u,t,i,k
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   runtest.sh of /kernel/filesystems/general/pjd-fstest
+#   Description: Tests POSIX features of file system
+#   Author: Zorro Lang <zlang@redhat.com>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+#   Copyright (c) 2019 Red Hat, Inc. All rights reserved.
+#
+#   This copyrighted material is made available to anyone wishing
+#   to use, modify, copy, or redistribute it subject to the terms
+#   and conditions of the GNU General Public License version 2.
+#
+#   This program is distributed in the hope that it will be
+#   useful, but WITHOUT ANY WARRANTY; without even the implied
+#   warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+#   PURPOSE. See the GNU General Public License for more details.
+#
+#   You should have received a copy of the GNU General Public
+#   License along with this program; if not, write to the Free
+#   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+#   Boston, MA 02110-1301, USA.
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+# Include rhts environment and beaker library
+. /usr/bin/rhts-environment.sh || exit 1
+. /usr/share/rhts-library/rhtslib.sh || exit 1
+
+
+# Test command, not the pipe (tee ...)
+set -o pipefail
+
+#prepare global values
+test -z "$TEST_PARAM_DEVICE" || TEST_DEV="${TEST_PARAM_DEVICE}"
+test -z "$TEST_PARAM_TESTMNT" || TEST_MNT="${TEST_PARAM_TESTMNT}"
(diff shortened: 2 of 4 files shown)


filesystems/general/pjd-fstest/README.md
```diff
@@ -0,0 +1,13 @@
+# pjd-fstest filesystem test suite
+Tests POSIX features of file systems. \
+Test Maintainer: [Zorro Lang](mailto:zlang@redhat.com)
+
+## How to run it
+
+### Dependencies
+Please refer to the top-leve README.md for common dependencies. Test-specific dependencies will automatically be installed when executing 'make run'. For a complete detail, see PURPOSE file
```
emanchado: *`top-level`

filesystems/general/pjd-fstest/runtest.sh
```diff
@@ -0,0 +1,254 @@
...
+	FSTYPES="$TEST_PARAM_FSTYPE"
+else
+	FSTYPES="ext3 ext4 xfs"
+fi
+FSTYPE=""
+TestsDir=""
+TmpDir=""
+
+# make sure the arguments all lower
+FSTYPE=$(echo ${FSTYPE}|tr [:upper:] [:lower:])
```
emanchado: Hm, shouldn't this be
```
FSTYPE=$(echo ${FSTYPES}|tr [:upper:] [:lower:])
```
In particular, that the echo should use `FSTYPES`, with `S` at the end. Otherwise, it's always printing an empty string, right?

filesystems/general/pjd-fstest/runtest.sh
```diff
@@ -0,0 +1,254 @@
...
+    fi
+
+    return 0
+}
+
+
+yum=$(select_yum_tool)
+
+check_supported_fs() {
+#	case ${FSTYPE} in
```
emanchado: Maybe delete this commented-out line?

filesystems/general/pjd-fstest/runtest.sh
```diff
@@ -0,0 +1,254 @@
...
+				rstrnt-abort --server $RSTRNT_RECIPE_URL/tasks/$TASKID/status
+				exit 0
+			fi
+		fi
+
+		if [[ $FSTYPE =~ ext ]] ; then
+			rlRun "/sbin/mkfs -t $FSTYPE -F $TEST_DEV $MKFS_OPTS" 0 "Creating filesystem on image"
+		else
+			rlRun "/sbin/mkfs -t $FSTYPE -f $TEST_DEV $MKFS_OPTS" 0 "Creating filesystem on image"
+		fi
```
emanchado: I wonder if it would be better that this whole `if` only had the flag (calculating if it's `-f` or `-F`) and then `rlRun` would be outside, using the flag calculated in the `if`...

filesystems/general/pjd-fstest/runtest.sh
```diff
@@ -0,0 +1,254 @@
...
+{
+	for testFeature in $TestsDir/*/; do
+		feature=`basename $testFeature`
+		rlPhaseStartTest "$feature: fs=\"$FSTYPE\" options=\"$FSOPTS\""
+		# prove in RHEL 5 and less does not support -f option
+                if [ $RHEL_DISTRO -ge 5 ] || rlIsFedora ;then
+			rlRun "prove -rf $testFeature 2>&1|tee output.log" 0 "Testing feature $feature"
+		else
+			rlRun "prove -rv $testFeature 2>&1|tee output.log" 0 "Testing feature $feature"
+		fi
```
emanchado: Same comment as above, for `prove`'s parameter.

filesystems/general/pjd-fstest/runtest.sh
```diff
@@ -0,0 +1,254 @@
...
+		rlFileSubmit "output.log" "output-$feature.log"
+		rlPhaseEnd
+	done
+
+	rlPhaseStartCleanup "Unmounted $TEST_DEV"
+		#rlRun "popd && popd" 0
```
emanchado: Delete?

filesystems/general/pjd-fstest/runtest.sh
ftjahjad: this is copied from original code, we could  change.  Trying to preserve the original code :-)

filesystems/general/pjd-fstest/runtest.sh
ftjahjad: this is copied from original code, we could  change it .  Trying to preserve the original code :-)

filesystems/general/pjd-fstest/runtest.sh
ftjahjad: correct, it should FSTYPES,  thanks for pointing out.

filesystems/general/pjd-fstest/runtest.sh
ftjahjad: correct, some left-over  comments  should be deleted.

ftjahjad:
cleaning up and fix logging result

filesystems/general/pjd-fstest/runtest.sh
```diff
@@ -0,0 +1,252 @@
...
+yum=$(select_yum_tool)
+
+check_supported_fs() {
+	local fstype=$1
+	case ${fstype} in
+		ext*|xfs)
+			:;;
+		*)
+			rlReport "Can't support $FSTYPE filesystem test" WARN
+			report_result $TEST WARN
```
veruu: Would be nice to have the indentation consistent with the commands below to make it more clear it's the same case

filesystems/general/pjd-fstest/runtest.sh
```diff
@@ -0,0 +1,252 @@
...
+				done
+			fi
+			TEST_DEV_LOOP=1
+			rlRun "dd if=/dev/zero of=test.img bs=1M count=500"
+			rlRun "losetup $TEST_DEV test.img"
+		fi
+
+		rpm -q xfsprogs
+		if [ $? -ne 0 -a $FSTYPE = "xfs" ]; then
+			${yum} -y install xfsprogs
```
veruu: Can this be added into test requirements instead of being installed here?

filesystems/general/pjd-fstest/runtest.sh
ftjahjad: good catch, I missed those  :+1:

filesystems/general/pjd-fstest/runtest.sh
ftjahjad: removing checking xfsprogs package. it is already in the test reqr makefile

PR #371, opened 2019-09-17
cpu/die: fix numa comparison

darcari:
As it turns out the index of numa node information provided by
lscpu is not determinate.  So instead, we can verify the numa
node data on a per index basis.

Secondarily, the calculation of logical die count didn't
always work correctly.  Instead of trying to determine if this
system is a beaker system, always check dmesg first and if no
information is available, check /var/log/messages.


darcari:
https://beaker.engineering.redhat.com/jobs/3791695

PR #373, opened 2019-09-19
Update storage test case to leverage cki_lib/libcki.sh

idorax:
Currently storage/include has its own library files (e.g. `libbkm.sh` and `libdebug.sh`). To make coding style consistent and save test code, it's time to leverage `cki_lib/libcki.sh` now.


idorax:
A basic unit testing is done on x86/rhel8, the job id is `J:3798416`.

storage/include/libstqe.sh
```diff
@@ -17,8 +17,7 @@
 # Boston, MA 02110-1301, USA.
 #
 
-source /usr/share/beakerlib/beakerlib.sh
-source $(dirname $(readlink -f $BASH_SOURCE))/libbkrm.sh
+source $(dirname $(readlink -f $BASH_SOURCE))/../../cki_lib/libcki.sh
```
emanchado: Out of curiosity: what's the difference between this and simply the following?
```
source ../../cki_lib/libcki.sh
```

storage/include/libstqe.sh
```diff
@@ -36,11 +35,12 @@ function stqe_init_fwroot
 
     # clone the framework
     typeset fwroot=$(stqe_get_fwroot)
-    rlRun "rm -rf $fwroot"
-    rlRun "git clone $STQE_GIT $fwroot" || rlAbort "fail to clone $STQE_GIT"
+    cki_run_cmd_neu "rm -rf $fwroot"
```
emanchado: Should this one really be the `_neu` variant? The original call doesn't have the `$BKRM_RC_ANY`...

storage/include/libstqe.sh
```diff
@@ -51,33 +51,33 @@ function stqe_init_fwroot
...
+    cki_pd
 
     return 0
 }
 
 function stqe_fini_fwroot
 {
     typeset fwroot=$(stqe_get_fwroot)
-    rlRun "rm -rf $fwroot"
+    cki_run_cmd_neu "rm -rf $fwroot"
```
emanchado: Another one that should probably be `_pos`?

storage/iscsi/params/runtest.sh
```diff
@@ -38,12 +36,13 @@ function cleanup
 function runtest
 {
     typeset fwroot=$(stqe_get_fwroot)
-    pushd "." && rlRun "cd $fwroot"
-    rlRun "stqe-test run -t iscsi/iscsi_params.py"
-    popd
+    cki_cd $fwroot
+    cki_run_cmd_pos "stqe-test run -t iscsi/iscsi_params.py" || \
```
emanchado: So, if this fails, the directory will be wrong because the `cki_pd` will not be run? Shouldn't that be avoided?

storage/include/libstqe.sh
idorax: `$(dirname $(readlink -f $BASH_SOURCE))/../../cki_lib/libcki.sh` is the absolute path, and `../../cki_lib/libcki.sh` is the relative path. It fails to load cki/libcki.sh if we use relative path. e.g.

- http://pastebin.test.redhat.com/799522

Hence, we have to use abolute path when sourcing shell script files.

storage/include/libstqe.sh
idorax: The original code was written by myself. I think it's better to remove directory $fwroot without checking its return code as we use `**rm -rf ...**`. So `cki_run_cmd_neu` is used instead of `cki_run_cmd_pos`.

storage/include/libstqe.sh
idorax: The same reason to L38.

storage/iscsi/params/runtest.sh
idorax: Good catch, updated. Please refer to:
* http://pastebin.test.redhat.com/799524

emanchado:
Should I just merge?

idorax:
Thank @emanchado for review again, please help to merge it.

PR #375, opened 2019-09-19
Workaround offline NFS servers

veruu:
We should skip the testcase for which the server is offline. But for now
we should at least SKIP the test instead of aborting. Yes, it's an infra
failure but it's reoccuring a lot an no testing is executed because the
recipe gets aborted early on.


veruu:
Discussed with @ftjahjad on IRC, merging

PR #377, opened 2019-09-23
networking: run netperf sctp_stream only when netperf support

linup2011:


linup2011:
> LGTM, although I'm not really that familiar with this code.
> 
> A question though: do we have to check IPv4 and IPv6 support separately? If not, maybe it would be nice to detect the support once, and then store that in a variable or something.

in theory, if sctp_stream is supported, ipv4 and ipv6 would both be supported. 
but there is another good point to detect twice, netperf can be checked once before really run it, which can avoid failure

rasibley:
Hi @emanchado you are welcome to merge changes if you feel confident they won't break things :-) I'll go ahead and merge since your off tomorrow, thanks for the feedback.

PR #378, opened 2019-09-23
ipsec/ipsec_basic_netns: extend the test time and send less tcp data

liuhangbin:
commit fbfe522fb40db9d6e1720a5c12d19fa7bfb8e365
Author: Xiumei Mu <xmu@redhat.com>
Date:   Wed Sep 11 16:37:52 2019 +0800

    ipsec/ipsec_basic_netns: extend the test time and send less tcp data

    This fix is for tcp connection fail on some mahines(Mustangs,aach64)

Signed-off-by: Hangbin Liu <liuhangbin@gmail.com>

### DIFF
--- a/networking/ipsec/ipsec_basic/ipsec_basic_netns/Makefile
+++ b/networking/ipsec/ipsec_basic/ipsec_basic_netns/Makefile
@@ -71,7 +71,7 @@ $(METADATA): Makefile
 	@echo "License:      GPLv2" >> $(METADATA)
 	@echo "TestVersion:  $(TESTVERSION)"	>> $(METADATA)
 	@echo "Description:  ipsec/ipsec_basic/ipsec_basic_netns">> $(METADATA)
-	@echo "TestTime:     20m" >> $(METADATA)
+	@echo "TestTime:     40m" >> $(METADATA)
 	@echo "Type:         Function" >> $(METADATA)
 	@echo "RunFor:       $(PACKAGE_NAME)" >> $(METADATA)
 # add any other packages for which your test ought to run here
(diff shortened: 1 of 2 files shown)


liuhangbin:
> Clueless question: are both changes needed? As in, if we send 10 times less data, maybe that's enough and we don't need to also extend the testing time?

I'm not sure, maybe sometimes the test case aborted due to time limit. I will let the author Xiumei answer this question.
> 
> Another thing: why is the commit message formatted like that? It seems generated, and having extra information that is probably not needed?

This is a backport from beaker repo. I quoted the commit id in description part, just like we backport patch from upstream. If you think this is not suitable, we can remove it.

liuhangbin:
> > Clueless question: are both changes needed? As in, if we send 10 times less data, maybe that's enough and we don't need to also extend the testing time?
> 
> I'm not sure, maybe sometimes the test case aborted due to time limit. I will let the author Xiumei answer this question.
>
OK, Xiumei said she forgot her acount.. She told me that these two parts are both fixing the same issue(test abort due to time limit), because sometimes the test would timeout during netperf testing.

veruu:
@liuhangbin thanks, does the test timeout change come with a kpet-db PR as well? Right now the timeout value in kpet-db overrides the one provided in makefile (@emanchado is working on getting rid of the duplication and the values will only be in kpet-db in the future), so this change will be ignored.

liuhangbin:
Ah, yes, thanks for the reminder. I will also update the kpet-db part.

emanchado:
So, should I merge this now, or wait for the kpet-db part?

veruu:
@emanchado it's ok to merge this one if you've reviewed it, no need to sync the changes ;)

PR #381, opened 2019-09-26
Add condition when installing RT dependencies

rasibley:
@veruu or @emanchado can you please review ? Note for openposix kernel-rt-devel is installed by default for RT jobs so it's not required to yum install. See J:3811529 for test results.

### DIFF
--- a/packages/perf/perftool-testsuite/runtest.sh
+++ b/packages/perf/perftool-testsuite/runtest.sh
@@ -222,8 +222,6 @@ rlJournalStart
 			which debuginfo-install || rlRun "$YUM -y install yum-utils dnf-utils" 0 "Installing {yum,dnf}-utils (it has not been present)"
 			which debuginfo-install # now it should be installed, but what if it fails...
 			if [ $? -eq 0 ]; then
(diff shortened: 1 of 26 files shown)


packages/perf/perftool-testsuite/runtest.sh
```diff
@@ -222,8 +222,6 @@ rlJournalStart
 			which debuginfo-install || rlRun "$YUM -y install yum-utils dnf-utils" 0 "Installing {yum,dnf}-utils (it has not been present)"
 			which debuginfo-install # now it should be installed, but what if it fails...
 			if [ $? -eq 0 ]; then
-				rlRun "debuginfo-install -y $KERNEL_PKG_NAME" 0 "Installing debuginfo for $KERNEL_PKG_NAME via debuginfo-install (it has not been present)"
```
rasibley: I removed this since it fails to install using the rt debuginfo package using debuginfo-install, the second statement always works in both cases.

rasibley:
@cki-bot please test

rasibley:
hi @veruu looks like it passed or at least any aborts were not related to the change tests.

PR #384, opened 2019-10-01
patch ltp dynamic_debug grep check

rasibley:
Now that we test debug kernels, the dynamic_debug01 test will fail the dmesg check when grepping for BUG, patch test to only check for BUG vs DEBUG in logs


rasibley:
I think so, I would just like @veruu feedback as well since she pointed this out to me.

PR #385, opened 2019-10-01
kpkginstall: allow different kernel-headers version

wizardofoz123:
With Fedora testing we may have different kernel and kernel-headers
package. This task expects the same version of these 2 packages, so
kernel-headers isn't installed. This patch tries to remedy that.


wizardofoz123:
The version difference happens mostly on rawhide and it's okay because of <devel reasons>.

wizardofoz123:
@cki-bot test

emanchado:
Only the last pipeline had errors, and only in s390x. Infra issue? This is the Beaker job: https://beaker.engineering.redhat.com/recipes/7427730#task100245659

wizardofoz123:
I've scan-read the results, including s390x. This should really be tested on Fedora, but it seems OK and doesn't break current streams. Merging.

PR #388, opened 2019-10-03
LTP workaround for oom aborting runtest.sh

rasibley:
Workaround for oom test killing runtest.sh, also fixing up another patch which used space indentation vs tabs, one more change related to a git clone option which would cause the checkout to sometimes fail, please see J:3823120 for a test job.

### DIFF
--- a/distribution/ltp-upstream/lite/runtest.sh
+++ b/distribution/ltp-upstream/lite/runtest.sh
@@ -58,7 +58,7 @@ function ltp_test_build()
 		pushd ltp; git pull  > /dev/null 2>&1; popd
 	else
 		if [ $GIT_USER_ADDR = 0 ]; then
-			git clone https://github.com/linux-test-project/ltp ltp  --depth=1\
+			git clone https://github.com/linux-test-project/ltp ltp \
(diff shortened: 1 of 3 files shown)


distribution/ltp-upstream/lite/runtest.sh
```diff
@@ -58,7 +58,7 @@ function ltp_test_build()
 		pushd ltp; git pull  > /dev/null 2>&1; popd
 	else
 		if [ $GIT_USER_ADDR = 0 ]; then
-			git clone https://github.com/linux-test-project/ltp ltp  --depth=1\
+			git clone https://github.com/linux-test-project/ltp ltp \
```
emanchado: Why remove the `--depth=1`?

rasibley: @emanchado I noticed that sometimes it would fail when using the `--depth=1` option, I saw the following in the logs` fatal: reference is not a tree: c9707b465a08397569920b676345474f44327200` I don't know why but removing it fixed it.

emanchado: Ah, spicy, "one of those"...

PR #389, opened 2019-10-08
Remove time limits from all tests

emanchado:
Remove "TestTime" from the old metadata files (testinfo.desc) created
by the Makefiles, and also remove "max_time" from new metadata files.

Ref. FASTMOVING-1201


emanchado:
@cki-bot test please!

PR #390, opened 2019-10-08
Add back cve tests for LTP

rasibley:
This PR is to re-enable cve tests in LTP, see beaker job 3829635 for results.


rasibley:
Thank you @emanchado I have updated the commit message to make it more clear.

PR #394, opened 2019-10-09
add tests for HW RAID for smartpqi and hpsa testing

mpatalan:
Tests created for HW RAID specifically for HPSA and SMARTPQI driver testing. Will be performed on specific servers (storageqe-29 - HPSA and storageqe-32 - SMARTPQI).


rasibley:
Hi @mpatalan I know we already did a thorough review while you were working on your PR, so I don't have any further comments here, the only question is where else is this supported besides rhel8, but we can find out soon once we test it in the pipeline. Thanks for the PR!

PR #399, opened 2019-10-10
Don't depend on the "service" command in tests

emanchado:
Switch calls to the "service" command to the BeakerLib equivalent
function. Because there's no rlServiceRestart, all of the "service xxx
restart" have been changed to two calls: "rlServiceStop xxx" and
"rlServiceStart xxx". Hopefully this doesn't change anything or break
any tests, but it's good to be aware that there could be a difference.

NB: The following files can be ignored because they are already using
systemctl when available, and only using service for older RHEL
versions:

* kdump/include/runtest.sh
* networking/route/pmtu/common/network.sh
* networking/firewall/000infralib/src/lib/service.sh
* networking/common/network.sh
* networking/common/service.sh
* networking/bridge/common/lib.sh


emanchado:
Hey yo, @cki-bot! Be a good boy and test this.

emanchado:
@cki-bot Attaboy!

networking/common/network.sh
```diff
@@ -136,7 +136,7 @@ reset_network_env()
 		pkill -f "nc -l"
 	
 		rsync -a --delete $networkLib/network-scripts.no_nm/ /etc/sysconfig/network-scripts/
-		 service network restart
+		service network restart
```
rasibley: @emanchado did you need to change this to use rlServiceStop and rlServiceStart ? Also I wonder if rlServiceRestore will do the equivalent ?

veruu:
I think it would make more sense to do `rlServiceStop X && rhServiceStart X` instead of having it on the separate lines. It would make it clean that the commands belong together. I don't think doing it this way would break anything since the command should finish before moving on the next one so it's just a suggestion.

emanchado:
@cki-bot Did you test this??

emanchado:
@veruu: I have made the suggested changes in a separate commit for easier review, if you want to have another look: https://github.com/CKI-project/tests-beaker/pull/399/commits/f889e90c0e649fde9b3b7304fb07775587082933

If you're cool with it and don't have any more comments, I'll squash these and merge :grin:

networking/common/network.sh
rasibley: @emanchado agreed I don't think it's necessary to create a new function, just curious if it was the same, sounds like it is indeed different based on the man page, it looks like you still have `service network restart` should this be changed to rlService ?

emanchado: No, `networking/common/network.sh` is one of the cases in which the test actually checks for the RHEL version (or the existence of `systemctl`), so this code is only run for old RHEL installations. The commit has the list of such files I found, to help investigating this in the future:

https://github.com/CKI-project/tests-beaker/pull/399/commits/b3ed0fa893e1da485c5ff0fb7079bafb56d597f8

rasibley: Ah ok thanks for confirming, I see that now :-)

PR #405, opened 2019-10-17
cpu/idle: resolve failure issues

darcari:
Two issues are causing failures.  One is counter rollover.  The
second is that the power utilization improvement on low power
systems does not meet the expectations of this script.  Fix the
first issue by handling rollover fix the second by returning SKIP.


darcari:
Tested here: https://beaker.engineering.redhat.com/jobs/3848008

rasibley:
ok I don't think it's a blocker from getting merged, thank you

PR #407, opened 2019-10-17
Add platform python and devel deps

ftjahjad:
Fix missing Python dependencies for storage/iscsi/params and storage/dm/common tests.


ftjahjad:
@rasibley ,
Thanks!!!

PR #410, opened 2019-10-18
vm/kvm-unit-tests: Allow KVM Unit Tests to run on ARK

mcondotta:
The QEMU binary is in a different location in Fedora and RHEL, updating the script to account for that.
Also updating the timeout for vmx_vmcs_shadow_test test since it still failed for Debug Kernel.

### DIFF
--- a/vm/kvm-unit-tests/runtest.sh
+++ b/vm/kvm-unit-tests/runtest.sh
@@ -137,6 +137,8 @@ fi
 # set the qemu-kvm path
 if [ -e /usr/libexec/qemu-kvm ]; then
     export QEMU="/usr/libexec/qemu-kvm"
+elif [ -e /usr/bin/qemu-kvm ]; then
+    export QEMU="/usr/bin/qemu-kvm"
 fi
 
 if [ -z "$QEMU" ]; then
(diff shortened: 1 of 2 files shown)


emanchado:
@mcondotta: I find the commit message a bit confusing. Shouldn't the PR description be the actual commit message? If so, can you change it?

In any case, can you update the branch? It seems out-of-date compared to `master`.

rasibley:
@emanchado there is a button (update branch) which allows up to rebase for contributors, which is pretty neat :-) I think the commit/PR description are correct IMO, we want to enable this test to support ARK kernels which runs on Fedora distro.

emanchado:
@rasibley: Thanks for the tip! I always doubt because I don't know if it's kind of bad form when doing it to someone else's branch (eg. I prefer rebasing) 😅

About the PR description: the PR description is not going to show up in Git history, right? Assuming that's the case, what I meant is that I would like to have the explanation in the PR description _also_ in the commit log, so that when someone checks the history they can see both things.

rasibley:
Hi @mcondotta would you mind adding your PR comment above to the commit body as @emanchado suggested, thank you!

PR #411, opened 2019-10-18
cpu/driver: handle older intel systems which do not have pstate support

darcari:
On some older intel systems the intel_pstate driver is not supported.
We expect those systems to be running acpi-cpufreq.  Avoid returning
FAIL in those cases.

### DIFF
--- a/cpu/driver/runtest.sh
+++ b/cpu/driver/runtest.sh
@@ -25,6 +25,25 @@ TMPDIR=/var/tmp/$(date +"%Y%m%d%H%M%S")
 
 source $CDIR/../../cpu/common/libutil.sh
 
+NON_PSTATE_PROCESSORS="26 31 46"
+skip_pstate()
+{
+    typeset model=$(lscpu | grep Model: | awk '{print $2}')
+    if [ -z "$model" ]; then
+	cki_log "unable to determine cpu model"
+	return 1
+    fi
+
+    for m in $NON_PSTATE_PROCESSORS; do
+	if [ $m -eq $model ]; then
+	    cki_log "model: $model - does not support intel_pstate"
+	    return 1
+	fi
+    done
+
+    return 0
+}
+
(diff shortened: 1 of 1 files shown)


darcari:
Tested here and here:

https://beaker.engineering.redhat.com/jobs/3850011
https://beaker.engineering.redhat.com/jobs/3849984

cpu/driver/runtest.sh
```diff
@@ -25,6 +25,25 @@ TMPDIR=/var/tmp/$(date +"%Y%m%d%H%M%S")
 
 source $CDIR/../../cpu/common/libutil.sh
 
+NON_PSTATE_PROCESSORS="26 31 46"
+skip_pstate()
```
emanchado: Minor suggestion: do you think that maybe this could be clearer if it was called `check_pstate_support` instead of `skip_pstate` (and flip the 0 and the 1s in the `return`s, of course)? I'm thinking that, usually, shell stuff uses 0 for "success" and non-zero for failure, so I was already confused, wondering if this function would use that convention, or a bool-like convention (as it seems to be using).

On top of that, having a negative ("skip") instead of a positive ("support") added a layer of complexity for me, especially since I wasn't sure if unknown models would skip pstate or not (in contrast, it's more or less intuitive that you should not assume that something is supported if you can't even detect what it is).

That said, it's a minor thing so feel free to ignore if you prefer it the way it is now.

darcari: Yes - I was using a bool like convention and I certainly didn't want 0 to be true.  That's probably more of a "C" thing.

I can make the changes you are suggesting.

emanchado:
Wait, why did you close it? 😕 Can you just push the new commits to the same branch, or force-push?

darcari:
I guess I should have done the force push, but I create a new branch instead.  I'll do that next time, but in this case since I've already started down the path - I prefer to keep going (retest is in progress).

PR #412, opened 2019-10-21
cpu/driver: handle older intel systems which do not have pstate support

darcari:
On some older intel systems the intel_pstate driver is not supported.
We expect those systems to be running acpi-cpufreq.  Avoid returning
FAIL in those cases.

### DIFF
--- a/cpu/driver/runtest.sh
+++ b/cpu/driver/runtest.sh
@@ -25,6 +25,19 @@ TMPDIR=/var/tmp/$(date +"%Y%m%d%H%M%S")
 
 source $CDIR/../../cpu/common/libutil.sh
 
+NON_PSTATE_PROCESSORS="26 31 46"
+check_pstate_support()
+{
+    for m in $NON_PSTATE_PROCESSORS; do
+	if [ $m -eq $1 ]; then
+	    cki_log "model: $model - does not support intel_pstate"
+	    return 1
+	fi
+    done
+
+    return 0
+}
+
 function verify_intel_cpufreq_driver
 {
     typeset driver=$1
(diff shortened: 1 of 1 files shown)


darcari:
v3: made changes suggested in v2

Test results here:

https://beaker.engineering.redhat.com/jobs/3852402
https://beaker.engineering.redhat.com/jobs/3852401

cpu/driver/runtest.sh
```diff
@@ -25,6 +25,25 @@ TMPDIR=/var/tmp/$(date +"%Y%m%d%H%M%S")
 
 source $CDIR/../../cpu/common/libutil.sh
 
+NON_PSTATE_PROCESSORS="26 31 46"
+check_pstate_support()
+{
+    typeset model=$(lscpu | grep Model: | awk '{print $2}')
+    if [ -z "$model" ]; then
+	cki_log "unable to determine cpu model"
+	return 1
```
rasibley: @darcari would you consider this a kernel bug or a setup issue, just wanted to confirm it should be a fail in this case ?

cpu/driver/runtest.sh
```diff
@@ -25,6 +25,25 @@ TMPDIR=/var/tmp/$(date +"%Y%m%d%H%M%S")
...
+    fi
+
+    for m in $NON_PSTATE_PROCESSORS; do
+	if [ $m -eq $model ]; then
+	    cki_log "model: $model - does not support intel_pstate"
+	    return 1
```
rasibley: This sounds like it's unsupported, should it return a skip ?

cpu/driver/runtest.sh
```diff
@@ -34,6 +53,18 @@ function verify_intel_cpufreq_driver
     typeset vendor=$(dmidecode -t 0 | grep Vendor: | \
                      cut -d: -f 2 | awk '{print tolower($1)}')
 
+    # make sure the given system supports p-state
+    check_pstate_support
+    if [ $? -ne 0 ]; then
+	# older systems do not support intel pstate
+	if [ $driver != "acpi-cpufreq" ]; then
+	    cki_log "intel (non-pstate) system is running: $driver"
+	    return $CKI_FAIL
```
rasibley: should we skip instead ?

cpu/driver/runtest.sh
darcari: I don't really think it is possible.

cpu/driver/runtest.sh
darcari: I we'd typically expect some cpu frequency driver to be running.  I guess they are all old systems so perhaps SKIP is better.

cpu/driver/runtest.sh
darcari: This is in check_pstate_support() which returns a 0 on success (meaning pstate is supported) and 1 on failure (meaning it is not).  So this is correct.

PR #414, opened 2019-10-24
FMK-1105 : powermgmt tests are intermittently failing in CKI

idorax:
The test cases are executed manually.
##### 1. Run test cases on intel bare-metal
* [cpupower/sanity](http://pastebin.test.redhat.com/808504)
* [rapl/powercap](http://pastebin.test.redhat.com/808510)

##### 2. Run test case by hacking has_kmod_intel_rapl() of libpwmgmt.sh to verify the case is skipped
* [rapl/powercap](http://pastebin.test.redhat.com/808512)


power-management/common/libpwmgmt.sh
```diff
@@ -0,0 +1,51 @@
...
+#
+
+source $(dirname $(readlink -f $BASH_SOURCE))/../../cki_lib/libcki.sh
+
+#
+# Check CPU vendor is Intel or not
+#
+function is_intel
+{
+    typeset vendor=$(grep vendor /proc/cpuinfo | sort -u | awk '{print $3}')
```
emanchado: Is the `sort -u` _only_ used to get the unique lines? If so, you can also use `uniq`, which is clearer in this context I'd say (because there's nothing to sort if they're all the same).

power-management/common/libpwmgmt.sh
```diff
@@ -0,0 +1,51 @@
...
+# Check kernel module 'intel-rapl' is available because power management
+# features require Xeon CPU
+#
+function has_kmod_intel_rapl
+{
+    # Always try to load kernel module 'intel-rapl' in case it is not loaded
+    cki_run_cmd_neu "modprobe intel-rapl"
+
+    # Check 'intel_rapl' has been loaded
+    cki_run_cmd_pos "lsmod | egrep '^intel.rapl '"
```
emanchado: Why `intel.rapl`? Isn't it always `intel_rapl`? Although I'm a bit confused about the fact that `modprobe` receives `intel-rapl` (with a dash)... Is that the reason, that we have to match both `_` and `-`?

power-management/common/libpwmgmt.sh
idorax: OK, use `uniq` instead of `sort -u`.

power-management/common/libpwmgmt.sh
idorax: I don't know why the module name is `inter-rapl` when using `modprobe`, but its name is `iner_rapl` when using `lsmod`. So I use '.' via `egrep` to match `_` in case in the future it is changed to be `-`. Anyway, I updated the function to use `_`.
```patch
diff --git a/power-management/common/libpwmgmt.sh b/power-management/common/libpwmgmt.sh
index 4049c53..c69b72a 100644
--- a/power-management/common/libpwmgmt.sh
+++ b/power-management/common/libpwmgmt.sh
@@ -24,7 +24,7 @@ source $(dirname $(readlink -f $BASH_SOURCE))/../../cki_lib/libcki.sh
 #
 function is_intel
 {
-    typeset vendor=$(grep vendor /proc/cpuinfo | sort -u | awk '{print $3}')
+    typeset vendor=$(grep vendor /proc/cpuinfo | uniq | awk '{print $3}')
     [[ $vendor == "GenuineIntel" ]] && return 0 || return 1
 }
 
@@ -46,6 +46,6 @@ function has_kmod_intel_rapl
     cki_run_cmd_neu "modprobe intel-rapl"
 
     # Check 'intel_rapl' has been loaded
-    cki_run_cmd_pos "lsmod | egrep '^intel.rapl '"
+    cki_run_cmd_pos "lsmod | egrep '^intel_rapl '"
     return $?
 }
```

PR #416, opened 2019-10-29
networking: increase robustness for the case

linup2011:
1. run sctp test when sctp is supported in netperf
2. check connection with ping in for loop
3. add -U for tcpdump


networking/tunnel/geneve/basic/runtest.sh
```diff
@@ -235,8 +236,10 @@ else
 			rlRun "tcpdump -r grev4.pcap -nnle | grep \"$SER_ADDR6.*> $CLI_ADDR6.*Geneve.*vni 0x4d2, proto TEB (0x6558).*$s_geneve1_mac > $c_geneve1_mac.*$gre_s_ip4 > $gre_c_ip4\""
 			[ $? -ne 0 ] && rlRun -l "tcpdump -r grev4.pcap -nnle"
 		fi
-		rlRun "$C_CMD tcpdump -i any -w grev6.pcap &"
+		rlRun "$C_CMD tcpdump -U -i any -w grev6.pcap &"
+		rlRun "sleep 5"
 		rlRun "$C_CMD ping6 $gre_s_ip6 -c 5"
+		rlRun "sleep 5"
```
emanchado: Yikes, that's lots of `sleep` calls :sob: Also, isn't 5 seconds extremely high?

networking/vnic/ipvlan/basic/runtest.sh
```diff
@@ -201,8 +201,26 @@ rlPhaseStartTest "local_netns"
...
+			done
+			rlRun "ip netns exec client ping 2.2.2.171 -c 2"
+			for i in {1..10}
+			do
+				if ip netns exec client ping6 2222::171 -c 1
+				then
+					break
+				else
+					sleep 1
+				fi
```
emanchado: Would it be possible to do something similar for the `sleep 5` above? In any case, but especially if we are to apply this kind of solution for the `sleep 5` calls above, would it be easy to make a function out of this? Something like `waitFor` giving `10` and the check command as parameters?

networking/tunnel/geneve/basic/runtest.sh
linup2011: on part systems like S390x, it would take long before tcpdump get ready, although I don't know why.

networking/vnic/ipvlan/basic/runtest.sh
linup2011: waitFor make sense, but it doesn't apply for the sleep 5 above, as I don't know how to check if tcpdump is ready of have finished capturing packets.
I would improve with waitFor and submit patchv2. and close this pull request.
thanks for your suggestion

PR #421, opened 2019-10-30
disable oom tests in ltp

rasibley:
The oom tests are causing the test harness to abort, removing exerimental patch as it's no longer needed and disabling the oom tests instead


rasibley:
See J:3868449 for test job

rasibley:
cc @jstancek

PR #431, opened 2019-11-12
Add required package dmidecode to Makefile because upstream doesn't have

idorax:


2019210654:
No, it doesn't break anything. If that package is installed (e.g. on RHEL8) by default, it's also okay. Please help to merge it, thanks!

PR #433, opened 2019-11-13
networking/tcp/tcp_keepalive: flush iptables before the test

xmmu:


networking/tcp/tcp_keepalive/runtest.sh
```diff
@@ -40,6 +40,9 @@ fi
 
 rlJournalStart
     rlPhaseStartSetup
+       rlRun "systemctl stop firewalld" 0-255
```
emanchado: Shouldn't this be `rlServiceStop firewalld` instead? That way it will be compatible with systems that don't have `systemctl`.

However, I don't remember what happens when stopping the service fails... but I don't know how important it is.

xmmu: Thanks for your review, I changed the code with  rlServiceStop, please check again.
Thanks
Xiumei

networking/tcp/tcp_keepalive/runtest.sh
```diff
@@ -40,6 +40,14 @@ fi
 
 rlJournalStart
     rlPhaseStartSetup
+       if [ $(rlGetDistroRelease) -ge "7" ]; then
+           rlServiceStop firewalld 0-255
+       else
+           rlServiceStop iptables 0-255
+           rlServiceStop ip6tables 0-255
```
emanchado: `rlServiceStop` interprets _all_ parameters as names of services to be stopped (see https://github.com/beakerlib/beakerlib/wiki/man#rlservicestop). Can you remove the `0-255` from the calls?

emanchado: Ah, and can you please update the branch, too?

xmmu: Please see the new patch, I removed 0-255, Thanks

PR #435, opened 2019-11-15
networking: not install iperf as not use in test

linup2011:
fail to install iperf sometimes,iperf is only used in route_fuzz_test
which would not run in cki,so not install it


linup2011:
> LGTM, although I wonder if it makes sense to remove it from the other four tests installing iperf. Surely the same logic (it fails to install sometimes, and that it doesn't run on cki) applies to the other cases, too?
> 
> Let me know if I should merge as-is, or if you want to have a look at the other tests and maybe make further changes before merging.

thanks for comment. I have submitted another patch based on your suggestion and devel's demand

linup2011:
> Rather than commenting so many lines, can we just delete them? I don't like having commented-out code, especially when it's so much.

I can delete them if needed

> 
> Also, would it make sense to remove the definitions of `iproute_upstream_install` and `netperf_install`, assuming all calls are removed now?

other networking related cases may need this, so just leave them here

> 
> Finally, can you please squash both commits into one?
I don't know how to do that. should I delete the pull request and submit another one?

emanchado:
Ok, delete the commented out lines, then, and leave the functions.

Never mind about squashing the commits, there's an option on GitHub and I can do it myself when merging this. But if you want to see how, look at the commands "git commit --amend" and "git rebase -i <commit-id>", they're both very useful in general so it might help you in other situations!

linup2011:
> Ok, delete the commented out lines, then, and leave the functions.
> 
> Never mind about squashing the commits, there's an option on GitHub and I can do it myself when merging this. But if you want to see how, look at the commands "git commit --amend" and "git rebase -i ", they're both very useful in general so it might help you in other situations!

done, thanks

PR #439, opened 2019-11-19
Update ltp to recent commit

rasibley:
Update LTP to checkout recent commit, update known issues file, also remove debug patch since this fix is already included upstream. See 3904066 for test results.


rasibley:
@veruu right :-) I added the beaker job id to the initial description, all green!

PR #440, opened 2019-11-20
enable ipmi for aarch64 and power

rasibley:
Updated both ipmi tests to run for both aarch64 and powerpc systems, depending on the ipmi implementation some commands and drivers are not available, tweaked the test to be more forgiving in these situations. Also to skip properly for systems which are not IPMI capable. The powerpc systems do not use dmidecode, however all bare metal powerpc systems are ipmi capable so we can ignore this check.


rasibley:
See J:3907377, J:3905176, J:3905171 for test jobs. Once merged, I'll update kpet to enable these arches and move testing to bare metal recipe.

PR #441, opened 2019-11-21
Try to install python3 if it's not present on server

subaf:


storage/include/libstqe.sh
```diff
@@ -20,7 +20,7 @@
 source $(dirname $(readlink -f $BASH_SOURCE))/../../cki_lib/libcki.sh
 
 STQE_GIT="https://gitlab.com/rh-kernel-stqe/python-stqe.git"
-STQE_STABLE_VERSION=${STQE_STABLE_VERSION:-"f1ddf69b"}
+STQE_STABLE_VERSION=${STQE_STABLE_VERSION:-"stable"}
```
emanchado: Is this change on purpose? The commit or the PR description don't mention anything...

storage/include/libstqe.sh
```diff
@@ -51,8 +51,11 @@ function stqe_init_fwroot
     for cmd in python python3 python2; do
         $cmd -V > /dev/null 2>&1 && python=$cmd && break
     done
-    [[ -n $python ]] || cki_skip_task "python not found"
-
+    if [[ -z $python ]]; then
+        cki_run_cmd_pos "yum install -y python36 python3-pip python3-setuptools" || \
+            cki_abort_task "fail python not found"
```
emanchado: Why `fail python not found` instead of simply `python not found`? Is it that you want to search for the string `fail`? If so, and if case sensitiveness doesn't matter, maybe it would look nicer/clearer if it read `FAIL: python not found`. If not, maybe drop the `fail` altogether?

storage/include/libstqe.sh
subaf: My bad. I forgot to mention it.

storage/include/libstqe.sh
subaf: I'm trying to keep fail messages formatting the same through the whole script but this reformatting of messages can be the next commit. Yes, we are using these fail messages to better track the failures.

PR #446, opened 2019-12-10
serge-testsuite: replace test script with a wrapper

WOnder93:
We recently moved the beakerlib wrapper for the upstream selinux-testuite to a [public repo](https://src.fedoraproject.org/tests/selinux/blob/master/f/kernel/selinux-testsuite), which should serve as a common source for all users of this test (currently RH internal repo and CKI).

This commit replaces the duplicated scripts with a thin wrapper that simply downloads the current version of the beakerlib script from the public repo and runs it.

Pros:
 * CKI will get test bugfixes & improvements faster,
 * no more test code duplication,
 * no need for us to open PRs to sync the code.

Cons:
 * changes to the test itself will no longer be in control of the CKI team,
 * test won't run on RHEL-5 (no big loss there...).

I tested the new test by cloning a couple CKI jobs and changing the fetch URL to my fork and branch and it seems to work as expected.

Overall, I believe this change will make things better for both sides. Let me know what you think.

(For more details see the commit message.)


rasibley:
@WOnder93 nice! I hope to see more of this with other beaker tasks to avoid maintaining two separate repos. Does the Fedora dist-git include the fixes for the sctp failures we were seeing in CKI ? Would you mind pasting the beaker job ID here so I can review the test job from your branch ?

WOnder93:
> @WOnder93 nice! I hope to see more of this with other beaker tasks to avoid maintaining two separate repos.

Glad you like it :) I have been pondering about doing something like this for some time, but it was @bachradsusi's initiative in our team that gave me the nudge to do it.

> Does the Fedora dist-git include the fixes for the sctp failures we were seeing in CKI ?

Not yet, I wanted to make sure the transition to public repo goes smooth before adding new changes, but I have a potential mitigation for it ready so expect it to land there soon.

> Would you mind pasting the beaker job ID here so I can review the test job from your branch ?

Actually, I tried just two: 3946008 (x86_64; this one failed due to missing kernel-devel files of the mainline kernel install, which is a separate issue that I forgot to mention) and 3946200 (ppc64le; passed). The output is practically the same, just a few extra lines that log what the wrapper is doing (there is also a dump of the dist-git commit ID, so you can trace back which version of the test was run).

Do you have any advice on the missing kernel-devel content? Can CKI install the files needed for a kernel module build also with mainline kernels or will we have to skip the module_load subtest when it's not there?

rasibley:
Thanks for the test job and I see the ticket filed internally, I pinged @major in the jira ticket, lets see what his response is, otherwise we may need to skip the subtest for upstream kernels.

WOnder93:
The outcome is that this is a known issue for now, so I patched the test to skip the module_load subtest when `/lib/modules/$(uname -r)/build/Makefile` dir is missing:
https://src.fedoraproject.org/tests/selinux/c/50a1ccb0d5edabe118cba7550632761681e602d6?branch=master

WOnder93:
(Moved `git rev-parse --verify` after `git checkout` so that `WRAPPER_GIT_BRANCH` works for branches other than the default branch.)

PR #450, opened 2019-12-16
audit-testsuite: on RHEL <8.1 skip filter_saddr_fam test

The-Mule:


rasibley:
Thanks @The-Mule, looks like you also updated kernel-fips-mode test to remove the `/var/tmp/fips-disabled`file, was this intended ? If so, would you mind updating the commit message to make it more clear.

The-Mule:
@rasibley, actually no. That was not intentional. I'll fix that.

The-Mule:
Cleaned.

rasibley:
Great thanks @The-Mule  ! LGTM !

PR #452, opened 2019-12-17
tracepoints/operational: add gcc check for retpolines

jbastian:


tracepoints/operational/runtest.sh
```diff
@@ -182,6 +182,32 @@ if [ $? -eq 0 ]; then
...
+if [ -e /sys/devices/system/cpu/vulnerabilities/spectre_v2 ]; then
+  CFLAGS=""
+  if [ "`uname -i`" = "x86_64" ]; then
+    CFLAGS="-mindirect-branch=thunk-extern"
+    CFLAGS+=" -mindirect-branch-register"
+    MITIGATION="retpoline"
+  elif [ "`uname -i`" = "s390x" ]; then
+    CFLAGS="-mindirect-branch=thunk-extern"
+    CFLAGS+=" -mindirect-branch-table "
+    CFLAGS+=" -mfunction-return=thunk-extern "
```
rasibley: Thanks @jbastian ! Is the space needed at the end here ? Also do you have a test job ?

jbastian: Oops, I see I put both leading and trailing spaces. It doesn't hurt, but it does look ugly. I'll clean that up.

Test job:
https://beaker.engineering.redhat.com/jobs/3966563

jbastian:
Patch is updated with trailing whitespace removed from those two lines.

tracepoints/operational/runtest.sh
jbastian: The patch is updated with the trailing whitespace removed from those two CFLAGS lines.

rasibley: @jbastian it was total nit-pick, looks good! Thanks for the test jobs to see it in action :-)

PR #453, opened 2020-01-09
fs/xfs/xfstests: blacklist MAX_LOCKDEP_CHAIN_HLOCKS

jencce:
For dmesg checking, this should mute the false alarm.

At the same time, bring generic/374 back.

Signed-off-by: xiangcai <jencce2002@gmail.com>

### DIFF
--- a/filesystems/xfs/xfstests/known_issues
+++ b/filesystems/xfs/xfstests/known_issues
@@ -1,7 +1,6 @@
 generic/466
 generic/084
 generic/139
-generic/374
 generic/394
 generic/471
 generic/427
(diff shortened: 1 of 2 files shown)


rasibley:
Thanks @emanchado for reviewing. This is  to help filter out a known failures when running on debug kernels, @jencce we actually need /usr/share/rhts/*strings to be present which I think is installed by rhts-test-env. However, I'm seeing conflicts when trying to install for both rhel7/rhel8, I cc'd you.

jencce:
Rachel Sibley <notifications@github.com> 于2020年1月9日周四 下午11:51写道：

> Thanks @emanchado <https://github.com/emanchado> for reviewing. This is
> to help filter out a known failures when running on debug kernels, @jencce
> <https://github.com/jencce> we actually need /usr/share/rhts/*strings to
> be present which I think is installed by rhts-test-env. However, I'm seeing
> conflicts when trying to install for both rhel7/rhel8, I cc'd you.
>
Yeah, this PR may not work.

Filing another PR.

Thanks!

PR #457, opened 2020-01-12
Ignore unsupported ipmi checks

rasibley:
IPMI supported checks vary widely depending on bmc implementation, also whether the system has updated firmware. For kernel testing, we want to verify the kernel doesn't get wedged or failures present in the logs, so ignore failures for all ipmi checks in this case.

### DIFF
--- a/ipmi/stress/ipmitool-loop/runtest.sh
+++ b/ipmi/stress/ipmitool-loop/runtest.sh
@@ -58,20 +58,20 @@ rlJournalStart
     # Execute various ipmitool commands in a loop
     for i in $(seq 0 10); do
         if [[ $(uname -m) != "ppc64le" ]]; then
-            rlRun "ipmitool sel clear"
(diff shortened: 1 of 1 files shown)


rasibley:
The message varies depending on the bmc implementation, here is an example:
```
: [ 19:57:00 ] :: [  BEGIN   ] :: Running 'ipmitool mc guid'
IPMI command failed: Invalid command
:: [ 19:57:00 ] :: [   FAIL   ] :: Command 'ipmitool mc guid' (Expected 0, got 1)
```
However the error varies depending on what machine your run against, I would have to maintain a file of all supported commands based on machines running updated firmware and ipmi supported functions, and even then if bugs are found they would be filed against the firmware or ipmitool, not the kernel. So to avoid this test from failing against kernel patches, I'm making it more generic. For kernel, it will fail if there is a IPMI driver failure found in the logs or if the system crashes after running ipmitool in a loop.

PR #458, opened 2020-01-17
FMK-1466 Include snippet of LTP failure (cont)

idorax:
Enable `distribution/ltp-upstream/lite/runtest.sh` to upload filtered test log to beaker.

### DIFF
--- a/distribution/ltp-upstream/include/runtest.sh
+++ b/distribution/ltp-upstream/include/runtest.sh
@@ -137,13 +137,20 @@ RprtRslt ()
     TEST=$1
     result=$2
 
+    logfile_run=$OUTPUTDIR/$TEST.run.log
+    logfile_fail=$OUTPUTDIR/$TEST.fail.log
+
+    # Always upload parsed test log for those failed test cases
+    GetFailureLog $logfile_run > $logfile_fail
+    [ -s $logfile_fail ] && SubmitLog $logfile_fail
+
     # File the results in the database
     if [ "$result" = "PASS" ]; then
         # I want to see the succeeded running log as well
-        SubmitLog "$OUTPUTDIR/$TEST.run.log"
+        SubmitLog $logfile_run
         report_result $TEST $result
     else
-        SubmitLog "$OUTPUTDIR/$TEST.run.log"
+        SubmitLog $logfile_run
         score=$(cat $OUTPUTDIR/$RUNTEST.log | grep "Total Failures:" |cut -d ' ' -f 3)
         report_result $TEST $result $score
     fi
--- a/distribution/ltp-upstream/include/ltp_log_parser.py
+++ b/distribution/ltp-upstream/include/ltp_log_parser.py
@@ -4,11 +4,32 @@
 import sys
 import getopt
 from itertools import groupby
+import re
 
 TC_START = "<<<test_start>>>"
 TC_END = "<<<test_end>>>"
 TC_NAME_FLAG = "tag="
-TC_RESULT_FAIL_FLAG = "TFAIL"
+TC_RESULT_FLAG = r'termination_type=exited\s+termination_id=\d+'
+#
+# XXX: Right here we define some result code according to source file:
+# https://github.com/linux-test-project/ltp/blob/master/include/tst_res_flags.h
+# /* Use low 6 bits to encode test type */
+# #define TTYPE_MASK	0x3f
+# #define TPASS	0	/* Test passed flag */
+# #define TFAIL	1	/* Test failed flag */
+# #define TBROK	2	/* Test broken flag */
+# #define TWARN	4	/* Test warning flag */
+# #define TINFO	16	/* Test information flag */
+# #define TCONF	32	/* Test not appropriate for configuration flag */
+# #define TTYPE_RESULT(ttype) ((ttype) & TTYPE_MASK)
+#
+TTYPE_MASK = 0x3f
+TPASS = 0
+TFAIL = 1
+TBROK = 2
+TWARN = 4
+TINFO = 16
+TCONF = 32
 TC_LOG_OUTPUT_INDENT = 4
 TC_LOG_SUMMARY_WIDTH = 79
 TC_LOG_OUTPUT_RAW = False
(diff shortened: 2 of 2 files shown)


distribution/ltp-upstream/include/runtest.sh
```diff
@@ -358,3 +373,11 @@ GetFailureLog ()
     local parser=$thisdir/ltp_log_parser.py
     python3 $parser -F -t 0 $logfile
 }
+
+GetAllLog ()
+{
+    local logfile=${1?"*** log file ***"}
```
emanchado: I don't understand this, what does it do? At first I thought it was `${1:?...}`, but it's not. This seems to... complain if `$1` is not set? Or something?

Anyway, once this is clarified or resolved, should I merge?

idorax: Hi @emanchado, yes, `${1?"*** log file ***}` is to make sure a log file is passed as $1.
I'm asking @rasibley to take a look at the commit, if she also thinks it's okay, please help to merge it, thanks!
BTW, for the test results, please review the Jira ticket `FASTMOVING-1466`.

idorax:
Hi @emanchado and @rasibley, python script `ltp_log_parser.py` cannot handle test cases failed without string 'TFAIL', so I fixed it. For more, please take a look at the comment I added to the Jira ticket.

rasibley:
Hi  @idorax looking at the test job, not all logs only three ltp subsystems are run (dio, run, mm) do you know why that is ? All LTP tests should run with each .run.log uploaded, I also noticed you uploaded each .all.log which isn't needed, only testarea.fail.log should be uploaded which includes the snippet failure.

idorax:
Hi @rasibley, for the job (`4024575`) you looked at which only have three ltp subsystems tested, it is because I set env `RUNTESTS=dio mm math`.  And .all.log is uploaded because I have set env `_DEBUG=true`, which is to test the usage of `SubmitLog` works fine. 
* https://github.com/idorax/tests-beaker/blob/dev.0107.FASTMOVING-1466.cont/distribution/ltp-upstream/include/runtest.sh#L148
```bash
# Upload the whole test log parsed if debug is on
if [ $_DEBUG = "true" ]; then
    GetAllLog $logfile_run > $logfile_all
    SubmitLog $logfile_all
fi
```
I can remove those lines if you think they should be removed.
BTW, a new job without setting _DEBUG and RUNTESTS is done, please take a look at `4048463`.

idorax:
@rasibley, here is the 'git diff' if we remove those lines to support '_DEBUG=true'.
* https://github.com/idorax/tests-beaker/commit/95e4af94707d11b364ff7ad263aa6939e22ea659

rasibley:
@idorax that's better, yes I would remove the debug statement as we only want to submit .run.log and fail.log if a failure occurred for that subsystem, I noticed there were multiple fail.logs uploaded in  your last job 4048463, looks like TCONF is being treated as a failure but actually TCONF means the test is unsupported, as you can see the test actually passed.

Test codes are described in this article: https://lwn.net/Articles/630542/
The result types are bit flags and the result is a bit field. The possible values are TPASS, TFAIL, TBROK, TWARN, and TCONF. TPASS and TFAIL are self-explanatory. TBROK means that something unexpected happened in the test setup and test was aborted. TWARN means that something unexpected happened but the test carried on. TCONF means that test is not suitable for the current configuration. The overall test exit value is composed of these bit flags. Note that TPASS is actually defined as zero so that the return value for successful test is also zero.

idorax:
Hi @rasibley, I have updated the fix according to your comments, the latest beaker job is `4052666`, please help to review again, thanks!

rasibley:
Thanks @idorax I tried to force a failure by removing some known issues, however I don't see the failure log uploaded this time ? See 4054333 for results. https://github.com/rasibley/tests-beaker/tree/log_parse_fail

I just want to be sure both pass and fail are doing the right thing, you can test out a failure by removing all the masked issues from the known issues file.

idorax:
Hi @rasibley, I looked into the test log of job `4054333`, the failed case is actually marked as 'TBROK', and currently our python script `ltp_log_parser.py`does filter 'TFAIL' only. Hence, I fixed it by taking both 'TFAIL' and 'TBROK' as **FAIL**. For details, please refer to:
- http://pastebin.test.redhat.com/833487
BTW, if a case has 'TWARN', the case should be regarded as **PASS** too?

idorax:
To test the latest fix, I hacked `distribution/ltp-upstream/lite/runtest.sh` to use my personal [**ltp** repo][02], which will fail a test case with 'TBORK'. Please review the beaker job `4055445`.

[02]: https://github.com/idorax/ltp/commit/23451f3e122594c241b3b7fa7ad9a80057dee05b#diff-ad197ce7200fb2b1e2c78eb0dbec0838

rasibley:
@idorax thanks! Getting closer :-) I don't see a .fail.log for ipc subsystem ? Also looks like the syscalls failure log included the known failure for msgstress04, not sure if we *should* still show known failures or better to skip them ? Maybe it's fine the way it is, I think TWARN are supposed to be ignored, e.g. not failures.  Maybe @jstancek can confirm ?

idorax:
Hi @rasibley,  for ipc.run.log, **'TWARN' is also regarded as 'FAIL' !**. Here is the details,
* http://pastebin.test.redhat.com/833924
Hence, we can have such a summary according to the test log from beaker jobs we have,
1. For 'TPASS', 'TINFO' and 'TCONF', the result is **PASS**,
2. For 'TFAIL', 'TBROK' and 'TWARN', the result is **FAIL**.
I've updated file ltp_log_parser.py accordingly, please take a look again, thanks!
```python
--- a/distribution/ltp-upstream/include/ltp_log_parser.py
+++ b/distribution/ltp-upstream/include/ltp_log_parser.py
@@ -180,7 +180,7 @@ def main(argc, argv):
     portion_flags = None
     for opt, arg in options:
         if opt in ("-F", "--fail"):
-            portion_flags = ["FAIL", "BROK"]
+            portion_flags = ["FAIL", "BROK", "WARN"]
         elif opt in ("-r", "--raw"):
             global TC_LOG_OUTPUT_RAW
             TC_LOG_OUTPUT_RAW = True
```

rasibley:
@idorax thanks for clarifying I didn't realize warn was actually treated as a failure, it all looks good now! However I'll wait to merge on Monday since I hate merging things on a Friday before the weekend :-)

PR #459, opened 2020-01-21
Make the jvm test more robust

emanchado:
The Maven call to install dependencies for the jvm test could fail due
to infrastructure problems. In those cases, the test should be aborted
so that infrastructure problems don't end up being marked as test
failures.


emanchado:
I had written an email to the maintainer, explaining the situation. However, I haven't been able to make a successful test run for the test (I wanted to do that first to make sure I hadn't broken anything). For example, see Beaker job J:4039338. Do you know what "new" means as a task status? The task seems to fail before running any tests, with "make: *** No rule to make target 'testinfo.desc'.  Stop."

Not sure how to proceed.

PR #460, opened 2020-01-21
Revert MAX_LOCKDEP_CHAIN_HLOCKS falsestrings

rasibley:
The dmesg falsestrings will now be set in the beginning of each recipe, this will apply to all beaker tasks. The MAX_LOCKDEP_CHAIN_HLOCKS string will be included as a falsestring. The current approach was causing false positives since the defaults were overwritten.


rasibley:
See J:4028620 for test results

rasibley:
@spbnick can you also review this please ? You can see it failed in the dmesg check since I removed the workaround, but I will merge the kpet-db MR at the same time to re-instate it at the beginning of the recipe.

spbnick:
LGTM!

PR #461, opened 2020-01-30
Make the jvm test more robust

emanchado:
The Maven call to install dependencies for the jvm test could fail due
to infrastructure problems. In those cases, the test should be aborted
so that infrastructure problems don't end up being marked as test
failures.

NOTE: this PR is only to use the PR bot that will run the tests, because I'm having trouble running the tests on Beaker (lack of Beaker-fu, most likely). The one that should probably be merged instead is https://github.com/CKI-project/tests-beaker/pull/459, which is already reviewed and approved.


emanchado:
@cki-bot Do your thing and test this, please! That was the whole point of this PR, after all.

PR #463, opened 2020-02-05
update ltp to recent commit

rasibley:
See 4054216 for test results


rasibley:
@emanchado please review :-)

PR #464, opened 2020-02-06
Make kpkginstall compatible with RHEL5

rasibley:
Older versions of bash don't support certain syntax, update cki_lib to make it compatible for rhel5. Default to yumdownloader command as --download option was added RHEL6 or later.

### DIFF
--- a/distribution/kpkginstall/runtest.sh
+++ b/distribution/kpkginstall/runtest.sh
@@ -313,7 +313,7 @@ function copr_prepare()
 function download_install_package()
 {
   # If download of a package fails, report warn/abort -> infrastructure issue
-  if $YUM install --downloadonly -y $1 > /dev/null; then
+  if $YUM install --downloadonly -y $1 > /dev/null || yumdownloader -y $1 > /dev/null; then
     cki_print_success "Downloaded $1 successfully"
   else
     cki_abort_recipe "Failed to download ${1}!" WARN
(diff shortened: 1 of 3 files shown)


rasibley:
@cki-bot please test

rasibley:
The last job included some bash error where the expected string wasn't displayed, use this one instead: J:4058344

distribution/kpkginstall/runtest.sh
```diff
@@ -466,6 +470,10 @@ else
       "${KVER}+debug"           # RHEL 8 style debug kernels
       "${KVER}.${ARCH}+debug"   # RHEL 8 style debug kernels
     )
+  elif grep -q "release 5" /etc/redhat-release; then
+    valid_kernel_versions=(
+      "${KVER_RHEL5}"
+    )
   else
     valid_kernel_versions=(
```
veruu: Stupid question, but why doesn't this match the RHEL5 release since it has both `${KVER}` and `${KVER}.${ARCH}`? The second one looks like it should match the original NVR before `sed` :thinking:

rasibley: I thought the same thing but it was still expecting ${KVER}.${ARCH} ?  had to strip the aarch using sed, but could have been something else causing it to fail, I can retry removing this to see.
```
✅ Found kernel version string in cache on disk: 2.6.18-440.el5.x86_64
ℹ️ Acceptable kernel version strings: 2.6.18-440.el5 
ℹ️ Running kernel version string:     2.6.18-440.el5
```

rasibley:
@veruu I now have the acceptable strings printed, it appears that the aarch was already included in KERN and was printed twice,  you can see the before and after output here: https://pastebin.com/eGG5paDS and test job:  4061489

I still have to strip the aarch for the rehl5 kernels otherwise it will abort as the KERN and running kernel do not match:
```
✅ Found kernel version string in cache on disk: 2.6.18-440.el5.x86_64
ℹ️ Acceptable kernel version strings: 2.6.18-440.el5.x86_64 
ℹ️ Running kernel version string:     2.6.18-440.el5
```

PR #465, opened 2020-02-08
update ltp to latest commit

rasibley:
@emanchado  can you review, we need to pull in one more fix, see 4061587 for test results.


rasibley:
@emanchado hmm I'm not sure, and I'm not able to reproduce, it may be a transient error and I think it's ok to merge. Thanks for the review!

PR #468, opened 2020-02-12
Add example xml running LTP

rasibley:
Add an example xml snippet for running LTP, this can be used as a base template for tests which run in beaker.


rasibley:
@veruu please review

PR #469, opened 2020-02-12
Add harness section for example.xml

rasibley:
Since we do rely on both beakerLib and restraint-rhts (soon to be removed), we should probably include the expected harness dependencies, will update to restraint vs restraint-rhts once legacy rhts commands
are removed.


rasibley:
@veruu one more please :-)

PR #470, opened 2020-02-13
WIP: Split jvm tests

rasibley:
The JVM test includes both DaCapo Benchmark and jcstress test suites. This test should be split into two cases, as they are distinguished tests suites and can hinder how we report our results in CI.


jvm/jcstress/PURPOSE
```diff
@@ -6,7 +6,3 @@ The Java Concurrency Stress tests (jcstress) is an experimental harness and a su
 to aid the research in the correctness of concurrency support in the JVM, class libraries,
```
emanchado: Same problem here in the initial paragraph.

jvm/dacapobench/PURPOSE
```diff
@@ -0,0 +1,7 @@
+The main purpose of this script is to ensure the kernel doesn't break java
+by verifying java is working appropriately through the following upstream
+test suites:
```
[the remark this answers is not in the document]
rasibley: Thanks I missed that! Should be fixed now.

jvm/jcstress/PURPOSE
rasibley: Should be fixed now.

rasibley:
See J:4067879 for test results, we need to wait to merge until kpet changes are merged first to change location.

rasibley:
ok I'm merging then, but this will be officially tested in kpet-db once I enable the suites.

PR #471, opened 2020-02-13
Update to latest stable version 20200120

rasibley:
Update our copy of stable LTP 20200120 to latest version to enable DCI testing for RHEL8 nightlies.

See J:4071803 for test results


emanchado:
Wait a minute, is the Beaker job correct? It seems to not have actually run any tests?

rasibley:
The beaker job is correct, the subsystem results are squashed into one file for the internal testing, you can expand and see results in RHELKT1LITE.FILTERED log. I will update the commit to be more clear. Note, this is is ltp vs ltp-upstream, this is difference between the two:

ltp-upstream is our copy which is very close to ltp github, we checkout a recent commit and update every 2 weeks so that we always pull in recent fixes and new tests. We use this for upstream kernel testing on Fedora distro.

ltp an open sourced version of what we run internally for RHEL. It grabs the latest stable release for LTP. We changed our internal jobs to just use the latest LTP beaker task since we don't cc upstream mailing lists, and helps keep us in sync with what we run internally. However DCI team (Distributed CI) would like to run the open sourced internal version at the partner lab for Dell, hence the reason for this PR, we need to update to a recent version since it hasn't been touched in a while.  This test happens to be running with an older version 20190115, but we need to update to latest version 20200120, stable versions are updated every ~3 months. https://github.com/linux-test-project/ltp/releases/tag/20200120

rasibley:
@emanchado I added one more change, and have updated the commit message, see test job J:4074467. I will be on PTO next week but please go ahead and merge if you approve :-)

emanchado:
LGTM

PR #474, opened 2020-02-25
Update lookaside location for iozone3 and dacapo artifacts

rasibley:
@veruu can you please review, see J:4091777 for test results.


veruu:
The test results look good! However, when I now check the URL locally it's not available? Maybe @inakimalerba moved it?

rasibley:
@veruu thanks! I changed the location and found one other test which was using it, new test job: J:4093331

PR #475, opened 2020-02-25
Update LTP to recent commit

rasibley:
See J:4093358 for test results.


rasibley:
@veruu please review, see test job: J:4099738, since s390x isn't yet enabled upstream I think it's ok to merge since it's still queued. I'll file a separate commit if the s390x deems it needs updates.

veruu:
@rasibley looks good! Agreed with the s390x, the queues are still pretty high right now. We might need to delay enabling it for upstream till they calm down. Feel free to merge when ready!

PR #478, opened 2020-03-03
kdump: Check and install grubby

emmawu:
grubby is not installed by default on F31.
To ensure crash memory is reserved properly on Fedora by updating kernel options by grubby, added a check and install grubby if it's not installed.

Verified on Beaker.

### DIFF
--- a/kdump/include/runtest.sh
+++ b/kdump/include/runtest.sh
@@ -370,7 +370,7 @@ PrepareKdump()
             # Kdump service will not be enabled if crashkernel=auto && system
             # memory is less the threshold required by kdump service.
             /bin/systemctl enable kdump.service || /sbin/chkconfig kdump on
-
+            rpm -q --quiet grubby || InstallPackages grubby
             Log "- Changing boot loader."
             {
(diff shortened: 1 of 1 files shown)


kdump/include/runtest.sh
```diff
@@ -370,7 +370,7 @@ PrepareKdump()
             # Kdump service will not be enabled if crashkernel=auto && system
             # memory is less the threshold required by kdump service.
             /bin/systemctl enable kdump.service || /sbin/chkconfig kdump on
-
+            rpm -q grubby || InstallPackages grubby
```
emanchado: Maybe better this way, so we don't add extra output? Or do you want the output to be there, to aid debugging?
```suggestion
            rpm --quiet -q grubby || InstallPackages grubby
```

emmawu:
@emanchado  Thanks for the suggestion! Updated the script.

emanchado:
Should I merge now?

rasibley:
LGTM ! Thanks @emmawu ! IMO it's ready to be merged.

PR #479, opened 2020-03-04
Update restraint and cki_lib environment part 1 of 2

ftjahjad:
fully updated test:
acpi    cki_lib    cpu           example      firmware  ipmi  kdump   memory
autofs  container  distribution  filesystems  iommu     jvm   legacy  misc

PR #481, opened 2020-03-11
CKI: create a test wrapper for upstream kernel selftest

liuhangbin:
Currently we only support bpf, net, forwarding, tc tests

Signed-off-by: Hangbin Liu <haliu@redhat.com>

### DIFF
--- a/kselftests/include.sh
+++ b/kselftests/include.sh
@@ -1,0 +1,151 @@
+#!/bin/bash
+[ ! "$JOBID" ] && rm -rf logs && mkdir logs && export LOG_DIR="$PWD/logs"
+
+if [ ! "$JOBID" ]; then
+	RED='\E[1;31m'
+	GRN='\E[1;32m'
+	YEL='\E[1;33m'
+	RES='\E[0m'
+fi
+
+new_outputfile()
+{
+	[ "$JOBID" ] && mktemp /mnt/testarea/tmp.XXXXXX || mktemp $LOG_DIR/tmp.XXXXXX
+}
+
+setup_env()
+{
+	# install dependence
+	# save testing environment
+	# export our new variable
+	export PASS=0
+	export FAIL=0
+	export OUTPUTFILE=$(new_outputfile)
+}
+
+clean_env()
+{
+	# clean environment
+	# restore environment
+	unset PASS
+	unset FAIL
+}
+
+log()
+{
+	echo -e "\n[$(date '+%T')][$(whoami)@$(uname -r | cut -f 2 -d-)]# " | tee -a $OUTPUTFILE
+	echo -e "\n[  LOG: $1  ]" | tee -a $OUTPUTFILE
+}
+
+submit_log()
+{
+	for file in $@; do
+		[ "$JOBID" ] && rstrnt-report-log -l $file || cp $file $LOG_DIR/
+	done
+}
+
+test_pass()
+{
+	echo -e "\n:: [  PASS  ] :: Test '"$1"'" >> $OUTPUTFILE
+	if [ $JOBID ]; then
+		rstrnt-report-result "${TEST}/$1" "PASS"
+	else
+		echo -e "::::::::::::::::"
+		echo -e ":: [  ${GRN}PASS${RES}  ] :: Test '"${TEST}/$1"'"
+		echo -e "::::::::::::::::\n"
+	fi
+}
+
+test_fail()
+{
+	SCORE=${2:-$FAIL}
+	echo -e ":: [  FAIL  ] :: Test '"$1"'" >> $OUTPUTFILE
+	if [ $JOBID ]; then
+		rstrnt-report-result "${TEST}/$1" "FAIL" "$SCORE"
+	else
+		echo -e ":::::::::::::::::"
+		echo -e ":: [  ${RED}FAIL${RES}  ] :: Test '"${TEST}/$1"' FAIL $SCORE"
+		echo -e ":::::::::::::::::\n"
+	fi
+}
+
+test_warn()
+{
+	echo -e "\n:: [  WARN  ] :: Test '"$1"'" | tee -a $OUTPUTFILE
+	if [ $JOBID ]; then
+		rstrnt-report-result "${TEST}/$1" "WARN"
+	else
+		echo -e "\n:::::::::::::::::"
+		echo -e ":: [  ${YEL}WARN${RES}  ] :: Test '"${TEST}/$1"'"
+		echo -e ":::::::::::::::::\n"
+	fi
+}
+
+test_pass_exit()
+{
+	test_pass $1
+	exit 0
+}
+
+test_fail_exit()
+{
+	test_fail $1
+	exit 1
+}
+
+test_warn_exit()
+{
+	test_fail $1
+	exit 1
+}
+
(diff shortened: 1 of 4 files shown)


kselftests/include.sh
```diff
@@ -0,0 +1,156 @@
...
+{
+	# clean environment
+	# restore environment
+	unset PASS
+	unset FAIL
+}
+
+log()
+{
+	#echo ":: [  LOG   ] :: : $1" | tee -a $OUTPUTFILE
```
emanchado: Can we remove commented-out lines?

kselftests/include.sh
```diff
@@ -0,0 +1,156 @@
...
+# Usage: run command [return_value]
+run()
+{
+	cmd=$1
+	# FIXME: only support zero or none zero, doesn't support 2-10, or 2,3,4
+	exp=${2:-0}
+	echo -e "\n[$(date '+%T')][$(whoami)@$(uname -r | cut -f 2 -d-)]# '"$cmd"'" | tee -a $OUTPUTFILE
+	# FIXME: how should we handle if there are lots of output for the cmd,
+	# and we only care the return value
+	eval "$cmd" &> >(tee -a $OUTPUTFILE)
```
emanchado: Oh, I had never seen this. How does it work?

kselftests/include.sh
```diff
@@ -0,0 +1,156 @@
...
+# Usage: watch command timeout [signal]
+watch()
+{
+	command=$1
+	timeout=$2
+	single=${3:-9}
+	now=`date '+%s'`
+	after=`date -d "$timeout seconds" '+%s'`
+
+        eval "$command" &
```
emanchado: Why the different indentation here, compared to the variables?

kselftests/include.sh
```diff
@@ -0,0 +1,156 @@
+#!/bin/bash
+# include beaker default environmnet
+[ -f /usr/bin/rhts_environment.sh ] && source /usr/bin/rhts_environment.sh
```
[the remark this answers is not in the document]
liuhangbin: Beaker still uses it. But yes, I can change it to beakerlib. I need to check how to use beaker lib to report pass/fail first.

kselftests/include.sh
liuhangbin: OK, I will

kselftests/include.sh
liuhangbin: This is used for redirecting the cmd output to log file while we still could get the return code.

kselftests/include.sh
liuhangbin: Sorry I didn't get your question. 

The run/watch functions are a re-write of rlRun and rlWatchdog. The purpose is to make user could run the test even without rhts/beakerlib.

kselftests/include.sh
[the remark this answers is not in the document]
liuhangbin: Hi @emanchado , I checked the rhts- functions and it looks there is no relation with the old RHTS. The current rhts- functions are wrappers of restraint functions. So I think it should be OK to use it. On the other hand, as my tests do not have rlPhase, I could use rlLog, rlReport or other beaker libraries separately. 

So may I keep using the rhts- functions?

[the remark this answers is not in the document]
emanchado: Sure thing! If you think it's fine, it's probably fine. I just wanted to make sure we were not using deprecated APIs. Thanks for checking!

kselftests/include.sh
emanchado: Oh, clever! I didn't know, that's pretty cool.

kselftests/include.sh
emanchado: I mean that the variables at the top of `watch` have an indentation of 6 spaces, but the `eval` and the rest of the commands have an indentation of 8 spaces. Shouldn't they be aligned?

emanchado:
BTW, I don't feel that great so I'll be taking today off. Maybe @rasibley can merge this if she approves?

kselftests/include.sh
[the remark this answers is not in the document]
rasibley: @emanchado is right, we are working on removing legacy rhts functions and libraries from CKI tests to make the transition to testing in VMs smoother, in addition to using other test frameworks like TMT. See https://bugzilla.redhat.com/show_bug.cgi?id=1802610 for more of a background.

You can see all the changes needed, but basically we need to do the following:
- Convert all rhts-* commands to rstrnt-* compatible commands:  https://restraint.readthedocs.io/en/latest/commands.html
- Remove any rhts libraries, e.g.` [ -f /usr/bin/rhts_environment.sh ] `
- Remove rhts-make.include and rhts-lint mentioned in makefile, instead set the variables:
https://github.com/CKI-project/tests-beaker/pull/479/files#diff-3817e1d835f842835666e91e0fbbb14eR37

Fendy was also referencing cki_lib/libcki.sh library, but I don't think it's needed unless you are specifically taking advantage of the the following variables in your test: (@ftjahjad  is that right ? )
https://github.com/ftjahjad/tests-beaker/blob/f903aecceaed8b604195a1d5f77195f40fa6eaf3/cki_lib/libcki.sh#L11

Fendy's PR for reference: https://github.com/CKI-project/tests-beaker/pull/479/files

rasibley:
@liuhangbin can't you combine this with the internal test wrapper ? Why do you need two separate tests ? I was hoping you would instead move your internal wrapper here and be able to test both RHEL and upstream kernels. Less work for you :-) Unless there is a specific reason for keeping them separate ?

kselftests/include.sh
liuhangbin: Ah, I see.  I will align them

kselftests/include.sh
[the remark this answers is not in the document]
ftjahjad: it is correct, most cki-beaker test use rhts-enviroment variables, 
https://github.com/ftjahjad/tests-beaker/blob/f903aecceaed8b604195a1d5f77195f40fa6eaf3/cki_lib/libcki.sh#L11 is setting beaker environment variables that deprecated  (rhts-environment.sh) in restrains.  
 
I think @rasibley  already covers everything, including  rhts-make.include and  rhts-<command> such as  replacing rhts-abort to   rstrnt-abort.

liuhangbin:
> @liuhangbin can't you combine this with the internal test wrapper ? Why do you need two separate tests ? I was hoping you would instead move your internal wrapper here and be able to test both RHEL and upstream kernels. Less work for you :-) Unless there is a specific reason for keeping them separate ?

@rasibley The most codes are copied from the internal test. And I did some improvement.  The reason I didn't use internal wrapper is there are some internal links. like kselftest rpms links.

rasibley:
> > @liuhangbin can't you combine this with the internal test wrapper ? Why do you need two separate tests ? I was hoping you would instead move your internal wrapper here and be able to test both RHEL and upstream kernels. Less work for you :-) Unless there is a specific reason for keeping them separate ?
> 
> @rasibley The most codes are copied from the internal test. And I did some improvement. The reason I didn't use internal wrapper is there are some internal links. like kselftest rpms links.

No problem, It should be possible to mask those internal links but we can do that later on, I don't want to block this. Thanks for all the improvements so far, I think it's ready to be merged, we can test it out further once we add it to kpet-db.

PR #484, opened 2020-03-17
ftrace-tracer: Test tracer availability by grep -w

sztsian:
Current logic of testing tracer availability can be mis-leading since
wakeup will also match wakeup_rt. It did not bring much hurt because now
wakeup_rt is always enabled with wakeup. However it worth to be fixed.

Reported-by: Philipp Rudo


emanchado:
Looks good, let's make @cki-bot test this before we merge.

PR #487, opened 2020-03-20
networking/ipsec: remove throughput test

xmmu:
Since CKI should be only running basic Tier0/Tier1 sanity level testing,
comment perf_test

Signed-off-by: Xiumei Mu <xmu@redhat.com>

### DIFF
--- a/networking/ipsec/ipsec_basic/ipsec_basic_netns/runtest.sh
+++ b/networking/ipsec/ipsec_basic/ipsec_basic_netns/runtest.sh
@@ -190,7 +190,8 @@ rlJournalStart
 
 	rlPhaseStartTest "$SUB_PARAM"
 		data_tests
-		perf_tests
+		# comment perf_tests for CKI-project test, as CKI should be only running basic Tier0/Tier1 sanity level testing
+		# perf_tests
 		cp /proc/crypto proc_crypto
 		rlFileSubmit proc_crypto
 	rlPhaseEnd


networking/ipsec/ipsec_basic/ipsec_basic_netns/runtest.sh
```diff
@@ -190,7 +190,7 @@ rlJournalStart
 
 	rlPhaseStartTest "$SUB_PARAM"
 		data_tests
-		perf_tests
+		#perf_tests
```
emanchado: Could you add a comment stating the reason why this is commented out? If I was reading the code and stumbled upon this line I would assume it was a leftover and my first reaction would be to delete it.

The alternative is actually deleting the line instead of commenting it out. Which is ok with me, but maybe leaving it along with an explanatory comment is clearer? I'm thinking that if someone ever noticed that the line is missing (eg. if they compare the code to the "upstream" version) maybe they will think the line was added afterwards and thus they would be improving the test by re-adding it...

What do you think?

rasibley: LGTM, maybe just add the comment as @emanchado and we can merge :-)

xmmu: I agree with you, and I updated the new patch , please help review. Thanks

PR #488, opened 2020-03-20
kdump: Install crash/kexec-tools from F32

emmawu:
1. Allow upgrading Kdump/Crash to latest F32 build
- This would hopefully solve the issue of userspace tools support on upstream kernels.
- Systemd/Dracut will be upgraded along with kexec-tools to avoid any service dependencies.
- Hardcode '--releasever=32' for now 

2. Add 2 options UPGRADE_FC_KDUMP/UPGRADE_FC_CRASH to
  enable/disable package upgarding.
- By default, both options are 'true'.
- Kexec-tools/Crash will not be upgraded to F32 build if UPGRADE_FC_KDUMP or UPGRADE_FC_CRASH is set to be false.
- This is just in case any regression introduced in latest F32 userspace builds.
- Applied to FC only.

3. Other minor fixes
- Kdump service is not restarted after reseting the default config
- Check if dmidecode is installed before generating bios log.
- Cleanup vmcores at the end of test
- Log the dracut/systemd pkg versions as well

### DIFF
--- a/kdump/include/runtest.sh
+++ b/kdump/include/runtest.sh
@@ -33,6 +33,9 @@ K_TMP_DIR="${K_TESTAREA}/tmp"
 K_REBOOT="${K_TMP_DIR}/KDUMP-REBOOT"
 C_REBOOT="./C_REBOOT"
 
+UPGRADE_FC_KDUMP=${UPGRADE_FC_KDUMP:-"true"}
+UPGRADE_FC_CRASH=${UPGRADE_FC_CRASH:-"true"}
+
 mkdir -p ${K_TMP_DIR}
 
 [[ "$FAMILY" =~ [a-zA-Z]+5 ]] && IS_RHEL5=true || IS_RHEL5=false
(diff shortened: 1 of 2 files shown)


kdump/include/runtest.sh
```diff
@@ -295,7 +298,7 @@ RhtsSubmit() {
 GetBiosInfo()
 {
     # Get BIOS information.
-    dmidecode >"${K_TMP_DIR}/bios.output"
+    which dmidecode && dmidecode >"${K_TMP_DIR}/bios.output"
```
idorax: Do we need to make sure package `dmidecode` is installed via Makefile? 
OR just keep the checking but don't upload file ${K_TMP_DIR}/bios.output if dmidecode is missing?

emmawu: I don't want to rely too much on beaker metadata/Makefile support. But you are right, I probably need to try installing the package first. Then decide whether to run dmidecode to generate bios.output. 
But still, the old code is harmless actually except generating an empty useless file. rhts-submit-log will not upload any empty file.
I will fix it.

emmawu: Hi @idorax, 
I fixed the dmidecode check. Can you please have a look?

idorax: Hi @emmawu, I didn't know that rhts-submit-log doen't upload empty file as a matter of fact, sorry for that. Your fix looks good to me, thanks!

rasibley:
Thanks Emma, LGTM!

PR #491, opened 2020-03-25
sleep after stopping firewalld

xmmu:
It's a continuous workaround for bz1755825
after stop firewalld service, sleep for slow host(s390x) or debug kernel

Signed-off-by: Xiumei Mu <xmu@redhat.com>


rasibley:
Thanks for the PR @xmmu !

PR #492, opened 2020-03-25
Audit testsuite rhel7 fix

The-Mule:
New tests were added recently to audit-testsuite (kernel audit upstream testsuite for audit used by this beaker tests). Two of them are not RHEL-7 compatible and need to be filtered out, they run correctly on RHEL-8.

### DIFF
--- a/packages/audit/audit-testsuite/runtest.sh
+++ b/packages/audit/audit-testsuite/runtest.sh
@@ -36,7 +36,7 @@ PACKAGE="kernel"
 GIT_URL=${GIT_URL:-"https://github.com/linux-audit/audit-testsuite.git"}
 
 # Optional test paramenter - branch containing tests.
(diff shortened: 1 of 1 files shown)


WOnder93:
I'd still suggest pinning to the current latest commit (ece04ca5517449930cafdddd8c076a258d0e0faf) so that when a new test is added upstream we don't immediately start getting a flood of CKI failures. We would then only bump the commit manually along with all needed sauce once we verify that the result works well across distros.

The-Mule:
Yeah, I forgot to take that approach (using the latest "stable" instead of master) from you last PR! :) Sorry about that. Let me fix that.

The-Mule:
Sure, sorry for that, it should be fixed now and it is ready to be merged. Thanks!

emanchado:
Excellent, thanks for the quick fix!

PR #494, opened 2020-03-30
audit-testsuite: bpf only for F>31 and RHEL>8.1

The-Mule:


The-Mule:
uh, oh, you are obviously right, let me fix that

emanchado:
Sure thing. I would have preferred a slightly clearer commit message (when compared to the patch), but it's just a nitpick so I'll merge.

PR #495, opened 2020-04-02
ipsec_baisc_netns: sleep longer for slow machines

xmmu:
Current sleep time is not enought on some slow machines like s390x,
so sleep double time.


xmmu:
> LGTM, but fix the typos in the commit summary (*`ipsec_basic_netns`) and in the commit description (*`enough`).

Sorry for the typos, I'll fix it.
> I always feel squishy when increasing sleep times. Isn't there some way to detect when the process is done, or at least apply that higher sleep only for s390x?
Not only s390x but also debug kernel of x86_64 is slow, and I think the initial time(1s) is too short even on fast host.
sleep longer is associated with better stability for the test case.

PR #496, opened 2020-04-03
Move static lookaside files to s3://cki-artifacts/lookaside/static

mh21:
We want to provide dynamic lookaside files like the beaker blacklist or the kernel fixes file. To that end, move the current files out of the way. This depends on the related deployment MR being merged first.


mh21:
Hi @cki-bot, could you test whether you can actually find those new files?

mh21:
http://lab-02.rhts.eng.bos.redhat.com/beaker/logs/tasks/108517+/108517159/taskout.log: 
lvm thinp sanity ok

PR #499, opened 2020-04-04
ltp: setup nr_hpage for hugetlb if MemAvailable is low

wangli5665:
This is mainly for getting rid of errors like:
  mem.c:822: BROK: nr_hugepages = 415, but expect 512

And also involves 'mem_low_mode' to avoid reporting false-positive(like memleak causes above error).

Signed-off-by: Li Wang <liwang@redhat.com>

### DIFF
--- a/distribution/ltp-upstream/lite/runtest.sh
+++ b/distribution/ltp-upstream/lite/runtest.sh
@@ -1,19 +1,7 @@
 #!/bin/bash
 
-# Copyright (c) 2011 Red Hat, Inc. All rights reserved. This copyrighted material 
-# is made available to anyone wishing to use, modify, copy, or
-# redistribute it subject to the terms and conditions of the GNU General
-# Public License v.2.
-#
-# This program is distributed in the hope that it will be useful, but WITHOUT ANY
-# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-# PARTICULAR PURPOSE. See the GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-#
-# Author: Zhouping Liu <zliu@redhat.com> 
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (c) 2020 Red Hat, Inc.
 
 . ../../../cki_lib/libcki.sh            || exit 1
 . /usr/share/beakerlib/beakerlib.sh	|| exit 1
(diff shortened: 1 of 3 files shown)


wangli5665:
@cki-bot test

emanchado:
Come on @cki-bot, don't be silly and test

distribution/ltp-upstream/lite/runtest.sh
```diff
@@ -1,19 +1,7 @@
...
-# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
-# PARTICULAR PURPOSE. See the GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program; if not, write to the Free Software
-# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-#
-# Author: Zhouping Liu <zliu@redhat.com> 
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (c) 2020 Red Hat, Inc.
```
emanchado, line 4: Is the author removed on purpose?

distribution/ltp-upstream/lite/runtest.sh
```diff
@@ -85,6 +74,72 @@ function ltp_test_build()
 	test_msg pass "LTP build/install successful"
 }
 
+function hugetlb_nr_setup()
+{
+       grep -q hugetlbfs /proc/filesystems || return
+       echo 3 >/proc/sys/vm/drop_caches
+
+       cat hugetlb.inc > hugetlb
```
emanchado, line 82: Why `cat >`, and not `cp`? Is `hugetlb` a special file in `/proc` or something? Because from this function it's not clear what the current directory is...

distribution/ltp-upstream/lite/runtest.sh
wangli5665: Yes, the author has left RedHat for more than 5 years. This test changes so much and no longer as it used to be, to avoid people send to the no-exist email account. So I remove it.

distribution/ltp-upstream/lite/runtest.sh
wangli5665: I didn't think too much for choosing 'cat' or 'cp', here just make a copy of the runtest file which we can hack it according to system available memory size.
After doing that, mv the hugetlb file to LTP installed directory and run with it.

wangli5665: and btw, you can find the config file at:
https://github.com/linux-test-project/ltp/blob/master/runtest/hugetlb

emanchado:
Should I merge as-is, then, once the tests finish and they pass (or fail for unrelated reasons)?

wangli5665:
These pipeline failures are unrelated to the pull request. From all test result, the ltp/hugetlb part works well:).

Btw, one potential error will be hit is caused by LTP itself bug, and I have sent another patch[1] to upstream ML. Fortunately, we didn't trigger it in these beaker jobs.
[1] http://lists.linux.it/pipermail/ltp/2020-April/016461.html

PR #508, opened 2020-04-15
Update libhugetlbfs setup warning to skip versus warn when hugepages are not available

idorax:


vm/hugepage/libhugetlbfs/Makefile
```diff
@@ -40,11 +40,12 @@ PATCH_DIR=patches
 ifndef SYSENV
 	SYSENV := $(shell uname -i)
 endif
-PYTHON_VERSION	:=$(shell which python3 &>/dev/null && echo python3)
+ARCH         := $(SYSENV)
+OS_MAJOR_RELEASE := $(shell grep -Go 'release [0-9]\+' /etc/redhat-release | sed 's/release //')
```
wangli5665, line 44: Shouldn't we consider compatible with Fedora and RHEL9? AFAIK, most of CKI jobs testing on Fedora+cki-kernel.

vm/hugepage/libhugetlbfs/dependency.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# For glibc-static dependence issue on rhel8
+# Bug 1567712 - glibc-static needs to be multilib on RHEL 8
+
+this_arch=$(uname -m)
+this_distro=$(grep DISTRO /etc/motd | cut -d '-' -f 3)
```
emanchado, line 7: It would be nice to have an example of `/etc/motd` (or at least `$this_distro`) in a comment. Reading this, I have no idea what `/etc/motd` usually contains (my local one is empty!) so it's hard to interpret what the rest of the code does.

vm/hugepage/libhugetlbfs/dependency.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if grep -q "release 8.*" /etc/redhat-release; then
+        if echo $this_distro | grep -q '.n'; then
+		choose=nightly
+        else
+		choose=rel-eng
+        fi
+        min_rel=$(grep -oE "[0-9]\.[0-9]" /etc/redhat-release | cut -d. -f2)
+
+	for ((; end_distro >= 0; end_distro-- )) do
```
emanchado, line 20: Can you please fix the indentation so it's consistent? The `if` block and the `min_rel` assignment are, for example, farther right than the `for` block.

vm/hugepage/libhugetlbfs/dependency.sh
```diff
@@ -0,0 +1,49 @@
...
+		echo "Trying option one: $build_root"
+		if curl --fail --head -s $build_root >/dev/null; then
+			success=1
+			break
+		fi
+	done
+
+	((success)) || build_root=http://download.eng.bos.redhat.com/rel-eng/rhel-8/RHEL-8/latest-RHEL-8.${min_rel}.0/compose/CRB/$this_arch/os/
+
+	echo Using repo $build_root
```
emanchado, line 31: Super nitpick: maybe use quotes in this `echo`, like in the one above? For consistency more than anything.

vm/hugepage/libhugetlbfs/dependency.sh
```diff
@@ -0,0 +1,49 @@
...
+skip_if_unavailable=1
+EOF
+
+	if [ $this_arch == x86_64 ]; then
+		rpm --quiet -q glibc-static || yum install -y glibc-static.x86_64 glibc-static.i686
+	else
+		rpm --quiet -q glibc-static || yum install -y glibc-static
+	fi
+
+	rm -fr /etc/yum.repos.d/buildroot.repo
```
emanchado, line 48: Hm, I'm curious about this: do we intend to delete the file we just created, or specifically `buildroot.repo`? Because the file created could have a different name, if `$repo_name` has a value...

And in any case, the `-r` is not needed, because it's always a file and not a directory?

idorax: Hi @emanchado, file `dependency.sh` is just copied from [cgit/tests/kernel](http://pkgs.devel.redhat.com/cgit/tests/kernel/tree/vm/hugepage/libhugetlbfs/dependency.sh). And the Makefile looks pretty complicated, I'll re-write it, please wait for a while ...

emanchado:
Another thing I just noticed: could you try to limit the commit summary (the first line) to ~50 characters? You can elaborate in the commit description, where you can also move the Jira ticket id if you want.

wangli5665:
@cki-bot

emanchado:
@cki-bot please test!

rasibley:
More changes came after the bot was triggered, @cki-bot please test once more

idorax:
@cki-bot please test!

emanchado:
The tests seem fine, except for one thing that I don't know:

Pipeline 535167 fails in aarch64 because `stress-ng` aborts (see Beaker recipe 8148911). It's probably unrelated, but can someone confirm?

idorax:
It is very time-consuming to run all test cases on all ARCH+TREE. It would be better if ckibot supports to test a single task because we only update a test case in many PRs.  That is, we should run all test cases only if the library (e.g. cki_lib) is updated, else we just run a single test case on all ARCH+TREE to save time as we just want to make sure the changes in PR works fine.

emanchado:
@idorax: can you add WIP to the title while you still work on it? Also, I'm not sure I understood your last comment. Do you mean you think it's fine and can be merged? Although we should run the tests again after the latest changes.

idorax:
Hi @emanchado, I noticed you helped to merge the master to current branch, thanks! Now I synced the branch via rebase to have only one commit.

idorax:
> @idorax: can you add WIP to the title while you still work on it? Also, I'm not sure I understood your last comment. Do you mean you think it's fine and can be merged? Although we should run the tests again after the latest changes.

WIP is added. Yes, I think stress-ng aborted failure is not caused by current commit because I don't update any library functions but case `vm/hugepage/libhugetlbfs` only. You know, to have a full run will take a lot of time, so it would be much better if ckibot supports to test a single task on all ARCH + TREE.

idorax:
@cki-bot please test again!

idorax:
Rebase is done and removed WIP.

emanchado:
The tests seem fine. It's mostly `check-install` failing. I'll merge.

PR #509, opened 2020-04-16
Delete unneeded php_mysql testing

ftjahjad:
We need to delete the php_mysql test  due to Php_mysql  kpet-db  testing is already deleted.


ftjahjad:
@emanchado 
Php_mysql test is already deleted from kpet-db. Based on  discussion  with @rasibley , we need to remove unneeded php-mysql test. 

@rasibley , 
do you have comments or ideas?

rasibley:
That's correct it was deleted a while back, but my bad I never deleted the test source here :-) See kpet-db MR 183

emanchado:
@rasibley: absolutely! I was just wondering 😅

PR #511, opened 2020-04-18
Revert getServerList() to avoid random failure

idorax:
After we removed getServerList() and defined CTHONSERVERS via kpet-db, the
    case fails randomly. Looks the client and nfs servers are not in the same
    domain, but we are not sure the root cause as a matter of fact. Hence, just
    go back to use getServerList() so as to save the effort to dig out the root
    cause. Note CTHONSERVERS is renamed to be _CTHONSERVERS to avoid updating
    kpet-db for the time being


idorax:
@cki-bot, please test!

filesystems/nfs/connectathon/runtest.sh
```diff
@@ -568,22 +585,16 @@ else
...
+if [ -n "$_CTHONSERVERS" ]; then
     # We can specify a list of servers it test against
     # For example, we can provide a customized list of
-    # servers CTHONSERVER="host1-nfs host2-nfs host3-nfs"
-    servers="$CTHONSERVERS"
+    # servers _CTHONSERVER="host1-nfs host2-nfs host3-nfs"
+    servers="$_CTHONSERVERS"
     echo " ========== Override server list =========="
-    echo " ***** INFO: CTHONSERVERS=$CTHONSERVERS *****"
+    echo " ***** INFO: _CTHONSERVERS=$servers *****"
```
emanchado, line 594: I wonder if it would be better to rename the variable only in the `if` condition, or maybe even not rename it there either, and simply add a `&& false` or similar at the end. It would be nice to add a comment above the condition, explaining that we're disabling that block temporarily.

idorax: Hi @emanchado, I use `if false; then ... fi` to disable the related code block and add some comments to explain why, please help to take a look again, thanks!

emanchado: Wait, is that what we want? If the `$CTHONSERVERS` variable *is* defined, won't `servers` end up empty? I was thinking of adding something to the condition to force it to always go to the `else` part...

idorax: Yes, you're right. Sorry for misunderstanding what you mean. Now I use
```bash
if false && [[ -n "$CTHONSERVERS" ]]; then
    ...
fi
```
instead.

emanchado:
@cki-bot: can you please test this again?

emanchado:
Pipeline 537884 has two failures when running the tests on x86_64 (Beaker job 154142) that I don't recognise:

1. LTP's syscalls fails (`sendmsg03` and `timerfd_settime02`; I think I have seen those fail recently...)
1. "audit: audit testsuite test" also fails, but it's hard for me to make sense of the output.

Can someone confirm if those are unrelated? In any case, can you please update the branch?

idorax:
Hi @em

> Pipeline 537884 has two failures when running the tests on x86_64 (Beaker job 154142) that I don't recognise:
> 
>     1. LTP's syscalls fails (`sendmsg03` and `timerfd_settime02`; I think I have seen those fail recently...)
> 
>     2. "audit: audit testsuite test" also fails, but it's hard for me to make sense of the output.
> 
> 
> Can someone confirm if those are unrelated? In any case, can you please update the branch?

The changes are not relevant to LTP and audit, can we just file ticket for them to record the failure? Anyway, current branch is rebased.

PR #513, opened 2020-04-21
Add libkcapi test for the AF_ALG interface

WOnder93:
This new test uses the libkcapi's upstream testsuite to test the AF_ALG
socket interface, which provides access to the kernel's Crypto API to
userspace. It is primarily intended to test libkcapi itself, but it has
proven invaluable in discovering kernel bugs as well.

NOTE: The test is currently failing on kernels 5.5+ due to a known issue:
https://lore.kernel.org/linux-crypto/CAAUqJDvZt7_j+eor1sXRg+QmrdXTjMiymFnji86PoatsYPUugA@mail.gmail.com/T/

### DIFF
--- a/packages/libkcapi/af_alg-test/README.md
+++ b/packages/libkcapi/af_alg-test/README.md
@@ -1,0 +1,18 @@
+# Libkcapi AF_ALG test
+
+Libkcapi test that exercises the AF_ALG interface and kernel Crypto API.
+
+Test maintainer: [Ondrej Mosnacek](mailto:omosnace@redhat.com)
+
+## How to run it
+Please refer to the top-level README.md for common dependencies.
+
+### Install dependencies
+```bash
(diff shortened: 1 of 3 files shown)


rasibley:
@WOnder93 thank you !! We are trying to remove legacy RHTS from this repo, would you mind instead using Restraint's metadata file instead ? https://restraint.readthedocs.io/en/latest/tasks.html#metadata

WOnder93:
> @WOnder93 thank you !! We are trying to remove legacy RHTS from this repo, would you mind instead using Restraint's metadata file instead ? https://restraint.readthedocs.io/en/latest/tasks.html#metadata

Hm, didn't know about that before... It sure looks nicer than the Makefile, I'll be happy to switch to it :)

> Also if you don't mind adding a simple README file as well, e.g. https://github.com/CKI-project/tests-beaker/blob/master/acpi/acpitable/README.md

Yes, will do.

rasibley (approved):
LGTM now! We can test it out further in kpet-db.

PR #515, opened 2020-04-22
audit-testsuite: run tests as unconfined_t

WOnder93:
This should work around a bug in the testsuite (to be fixed by [1]) that
makes it fail when run as system_u:system_r:unconfined_service_t:s0,
which is what the new restraint runs it with now.

[1] https://github.com/linux-audit/audit-testsuite/pull/96

Cc: @The-Mule


rasibley (approved):
LGTM, thank you !

PR #518, opened 2020-04-28
Update tests from legacy makefile to restraint metadata (01)

idorax:
This is the 1st PR to fix FASTMOVING-1755. Two test cases are updated,
* acpi/acpitable
* container/podman

A beaker job is done, the job id is `4225770`.


idorax:
@cki-bot, please test!

container/podman/README.md
```diff
@@ -0,0 +1,16 @@
+# Podman test suite
+Verify that podman can run integration tests with the system. \
+Test Maintainer: [Major Hayden](mailto:major@redhat.com)
+
+## How to run it
+Please refer to the top-level README.md for common dependencies.
+
+### Install dependencies
+```bash
+dnf -y install $(grep '^dependencies=' metadata | awk -F'=' '{print $NF}' | tr ';' ' ')
```
emanchado, line 10: Nice! This is going to be the same for every test that uses the modern `metadata` stuff, right? I wonder if there is a better way to standardise how to install the dependencies. It sucks a bit to have to include these instructions everywhere 🙁

idorax: Yes, will use `metadata` for all test cases. 
I'll file a script to install dependencies and update README.md, please wait for a while...

emanchado: It doesn't have to be in this PR BTW! I was just wondering, thinking out loud, and wondering about your opinion.

So we can merge as-is if you want, or I can wait for the script if you prefer.

ftjahjad:
@idorax , 
I am thinking not delete makefile , and just added metadata and point to use  restraint :bash ./runtest.sh

@rasibley 
do you have opinion to keep makefile or delete it?

container/podman/README.md
idorax: The script to install dependencies is added. Please take a look again, thanks!

idorax:
> @idorax ,
> I am thinking not delete makefile , and just add metadata and point to use restraint :bash ./runtest.sh
> 
> @rasibley
> do you have opinion to keep makefile or delete it?

@ftjahjad, I was thinking it is not very helpful to keep Makefile because:
1. With restraint metadata file, the most important part `repoRequires` for execute runtest.sh can be supported;
2. It's easy to install packages (e.g. git, patch, ...) required in setup phase when running runtest.sh; That is, if we don't use `dependencies=` in metadata file, it doesn't break anything as we can move the work (installing packages) to runtest.sh;
3. Keeping legacy Makefile may introduce maintenance noise. For those testers who don't know `Makefile v.s. metadata` very well, they may keep writing Makefile which is difficult to maintain because of bad readability, e.g. [vm/hugepage/libhugetlbfs/Makefile](http://pastebin.test.redhat.com/860035)

@rasibley, what do you think?

ftjahjad:
@idorax 
I totally agrees, fully restraint is very clean compared with legacy rhts.  Thinking to keep makefile for the tests are  already in cki test beaker  awhile and remove it later when we done fully testing . a new cki test-beaker should  uses restraint metadata.

@rasibley 
I thinks, we already had discuss  on keeping makefile  (keeping awhile, and remove it later when  tests are fully testing in pipelines on certain time), or delete makefile when working on updating to metadata.
Either way, I am ok :-). I proposed deleting makefiles initially for PR simplicity.

rasibley:
@ftjahjad @idorax if you can easily remove the Makefile, then please remove it. If it's more complicated to do so, we can do it in phases if necessary, some tests which utilize the Makefiles heavily like LTP for example. Otherwise metadata file is much cleaner and the preferred method to store test metadata in Restraint, so lets kill those Makefiles :-)

ftjahjad:
@rasibley
Thanks! 
LTP is the mother of makefile :-D. we  discussed it  before and  I totally agreed "doing  in phases" is a best approach.
Most makefiles  are easily removed.
 I updated some tests that require additional works due on nonsimple makefile in  jira ticket#1755

idorax:
Hi @rasibley and @ftjahjad, the conclusion is to keep Makefile in this PR and remove all Makefiles which can be removed in a separate PR after ticket 1755 is done, right? If yes, I'll bring the deleted Makefiles in this PR back soon.

emanchado:
Oh, I understood from Rachel's comment that we should remove the Makefiles ("if you can easily remove the Makefile, then please remove it. If it's more complicated to do so, we can [...]"). No sense in bringing them back if we have already deleted them?

In any case, the new script looks great, thanks! I'm ok either way, I think we can merge this with or without Makefiles. So if someone can confirm, I'll merge this. Or feel free to merge yourself, whoever has merge permissions.

idorax:
Hi @emanchado, thank you for your clarification. It is nice to remove Makefile while adding metadata file because I also updated runtest.sh by adding TEST var to it (Note TEST var was defined via Makefile). Please help to merge this PR after getting the feedback from Rachel, thanks again!

PR #520, opened 2020-05-06
Drop patches from amtu and use latest autoconf-amtu-1.1 branch

rasibley:
Recent changes were pushed to upstream amtu to fix gcc10 failures
for F32 migration. The master branch is now synced with the
autoconf-amtu-1.1 branch, so there is no need to apply patches.


rasibley:
See test job J:4239924 and related discussion in FMK-1798

rasibley:
Thank you :smile: !

PR #521, opened 2020-05-08
Update tests from makefile to metadata (02)

idorax:
This PR focuses on tests in networking, a beaker job `4248412` is done to test rhel8/x86_64.


idorax:
@cki-bot, please test!

networking/socket/fuzz/metadata
```diff
@@ -1,11 +1,12 @@
...
 license=GPLv2
 confidential=no
 destructive=no
 
 [restraint]
-entry_point=make run
-repoRequires=networking/common;cki_lib
+entry_point=bash ./runtest.sh
+dependencies=gcc;lksctp-tools-devel
+repoRequires=cki_lib
```
ftjahjad, line 12: This seems needing  networking/common  on repoRequires. 
The test runtest.sh  has code for sourcing networking/common:
. ../../common/include.sh

networking/bridge/sanity_check/metadata
```diff
@@ -0,0 +1,12 @@
...
+name=networking/bridge/sanity_check
+owner=Fei Liu <feliu@redhat.com>
+description=Networking bridge sanity test
+license=GPLv2
+confidential=no
+destructive=no
+
+[restraint]
+entry_point=bash ./runtest.sh
+dependencies=
```
ftjahjad, line 11: We could delete dependencies  in restraint entry if not used.
[restraint] entry is required  entry_point  minimally.

networking/firewall/000infralib/metadata
```diff
@@ -0,0 +1,13 @@
...
+owner=Networking QE
+description=Helper functions for networking firewall tests
+license=GPLv2
+confidential=no
+destructive=no
+
+[restraint]
+entry_point=bash ./runtest.sh
+dependencies=
+softDependencies=
```
ftjahjad, line 12: mentioned on previous comment,  we could  remove empty dependencies and softDependencies lines in the test metadata

networking/route/pmtu/metadata
```diff
@@ -0,0 +1,12 @@
...
+name=networking/route/pmtu
+owner=Jianlin Shi <jishi@redhat.com>
+description=Networking pmtu tests 
+license=GPLv2
+confidential=no
+destructive=no
+
+[restraint]
+entry_point=bash ./runtest.sh
+softDependencies=net-tools;tcpdump;nmap;ethtool;iproute;nc;nmap-ncat;socat;libtool;bind-utils;bridge-utils;libteam;teamd;traceroute;expect;pexpect;python3-pexpect;bc;wget;bzip2;gcc;git;patch;rsync;texinfo
```
ftjahjad, line 11: Does it mean dependencies instead of softDependencies?  :-)

networking/socket/fuzz/metadata
idorax: Updated.

networking/bridge/sanity_check/metadata
idorax: Removed.

networking/firewall/000infralib/metadata
idorax: Removed.

networking/route/pmtu/metadata
idorax: Since there are more than 20 packages are required, I don't know what packages will break the process of handling dependencies, so use `softDependencies` instead of `dependencies` to make sure the test can continue. You know, if a package listed in `dependencies` cannot be installed, it will abort in handling `metadata` phase. But if a package listed in `softDependencies` (e.g. bridge-utils) cannot be installed, it will not.

idorax:
> -1cry CKI pipeline failed [#521 (comment)](https://github.com/CKI-project/tests-beaker/pull/521#issuecomment-627194244)
> 
> https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/564033
> https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/564032
> https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/564031

[upstream](https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/564033) : BAD
[rhel7](https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/564032) : BAD
[rhel8](https://xci32.lab.eng.rdu2.redhat.com/cki-project/cki-pipeline/pipelines/564031): GOOD

@cki-bot, please test again!

networking/route/pmtu/metadata
ftjahjad: Yes,it is true. It is taking extra time (trials and errors) to know which packages  should be in  softDependencies.
I am thinking  these packages  should be in softDependencies: net-tools;bridge-utils;nmap-ncat;libteam;teamd;pexpect;python3-pexpect. 

@rasibley ,
should we need to put extra works to find packages that should be in sofDependencies metadata when involving many packages?
Either way, should be good.

rasibley: IMO yes we should take a look at the previous harness.log where we use Makefiles for supported releases to determine which packages didn't install. I'm hoping this isn't too painful, but we should try to do the right thing, otherwise this could lead to confusion if a non-rh contributor is running our tests.

idorax: To figure out what packages should be listed in `softDependencies`, I reserved two hosts (the 1st one's os is RHEL7, the 2nd one's os is RHEL8), then I had a try to install packages (e.g. `bc;bind-utils;bzip2;ethtool;expect;bridge-utils;pexpect;python3-pexpect`), if any one among them could not be installed on either RHEL7 or RHEL8, it should be listed in `oftDependencies`, e.g. `bridge-utils;pexpect;python3-pexpect`. Let's wait for the test results triggered off by cki-bot...

idorax:
@cki-bot, please test again!

idorax:
@cki-bot, please test again!

ftjahjad:
Looks good to me .

rebase required :-)

@rasibley , could you review again , Thanks :-)

networking/driver/sanity/runtest.sh
```diff
@@ -2,6 +2,8 @@
 
 . ../../../cki_lib/libcki.sh || exit 1
 
+TEST=$(egrep '^name=' metadata | awk -F'=' '{print $NF}')
```
rasibley, line 5: Hi @idorax is there a reason your adding this variable  here ? Seems like it's only needed for the common library ? Should we remove it here or other places where common lib is not a dep ?

rasibley:
LGTM! Just left one comment for @idorax

networking/driver/sanity/runtest.sh
idorax: Hi @rasibley , it is because `TEST` var was defined in `Makefile`, once Makefile is removed, I'd like to define it in all runtest.sh. For more about `TEST=` in Makefiles, please refer to:
o http://pastebin.test.redhat.com/867415

rasibley: TEST was just used to generate the Makefile metadata,  so I don't think it's needed unless the test is actually referencing that var elsewhere. Maybe cbouchar@redhat.com can confirm ? More details here:  https://restraint.readthedocs.io/en/latest/remove_rhts.html

``` 
export TEST=/examples/no-rhts/sample-before
export TESTVERSION=1.0

BUILT_FILES=

FILES=$(METADATA) runtest.sh Makefile PURPOSE

.PHONY: all install download clean

run: $(FILES) build
        ./runtest.sh

build: $(BUILT_FILES)
        test -x runtest.sh || chmod a+x runtest.sh

clean:
        rm -f *~ $(BUILT_FILES)

include /usr/share/rhts/lib/rhts-make.include

$(METADATA): Makefile
        @echo "Owner:           User ABC1 <userabc1@example.com>" > $(METADATA)
        @echo "Name:            $(TEST)" >> $(METADATA)
        @echo "TestVersion:     $(TESTVERSION)" >> $(METADATA)
        @echo "Path:            $(TEST_DIR)" >> $(METADATA)
        @echo "Description:     Sample-before-no-rhts" >> $(METADATA)
        @echo "Type:            Sanity" >> $(METADATA)
        @echo "TestTime:        5m" >> $(METADATA)
        @echo "Priority:        Normal" >> $(METADATA)
        @echo "License:         GPLv2+" >> $(METADATA)
        @echo "Confidential:    no" >> $(METADATA)
        @echo "Destructive:     no" >> $(METADATA)
        @echo "Releases:        -RHEL7 -RHEL8" >> $(METADATA)
        @echo "Architectures:   x86_64" >> $(METADATA)

```

idorax: Hi @rasibley, thanks for the clarification. I have removed `TEST=...` from all runtest.sh, please take a look again, thank you!

- http://pastebin.test.redhat.com/867485

rasibley: Thanks @idorax just trying to simplify things ,but lets have the bot test just in case I was wrong :-) @cki-bot please test once more.

rasibley:
I didn't see the bot triggered maybe because it was in a reply vs a top level comment, @cki-bot please test once more.

rasibley:
@idorax looks like Networking sctp-auth: sockopts test is consistently failing for rhel7/rhel8, otherwise testing looks good, see 4281033 for an example:
`:: [ 14:06:18 ] :: [ WARNING  ] :: rlReport: Only PASS/WARN/FAIL results are possible.`

I think this is the only test that was actually using $TEST :-) can you add it back just for this test ? 
https://github.com/CKI-project/tests-beaker/blob/master/networking/sctp/auth/sockopts/runtest.sh#L46

Also just provide one example test job afterwards for rhel7/rhel8, we don't need to trigger the bot for all testing again :-)

idorax:
> @idorax looks like Networking sctp-auth: sockopts test is consistently failing for rhel7/rhel8, otherwise testing looks good, see 4281033 for an example:
> `:: [ 14:06:18 ] :: [ WARNING ] :: rlReport: Only PASS/WARN/FAIL results are possible.`
> 
> I think this is the only test that was actually using $TEST :-) can you add it back just for this test ?
> https://github.com/CKI-project/tests-beaker/blob/master/networking/sctp/auth/sockopts/runtest.sh#L46
> 
> Also just provide one example test job afterwards for rhel7/rhel8, we don't need to trigger the bot for all testing again :-)

Hi @rasibley , I updated some `runtest.sh` by adding env TEST. For details, please refer to:
* http://pastebin.test.redhat.com/869481

Sometimes TEST is required in runtest.sh file, sometimes it is required in library functions which is more complicated as the calling stack is not easy to determined. Hence, if we can add TEST to all runtest.sh, it would be safer. What do you think?

idorax:
Hi @rasibley, the latest changes (i.e. adding env TEST) are tested, and test results looks good.
* RHEL8: 4286571
* RHEL7: 4286597

rasibley:
Thanks @idorax ! LGTM now :-)

PR #522, opened 2020-05-08
kdump: Added test cleanup/store and set KDUMP_FILE_LOAD to off 

emmawu:
A few changes
- Added steps to restore default kdump config/systconfig and cleanup vmcore. 
- Renamed RhtsSubmit to RstrntSubmit
- Do not uppload kdump initramfs img by default. It's for debugging purpose only.
- Set KDUMP_FIlE_LOAD to off. In Fedora, there is a feature change made in kdump that by default it will use kexec_load_file calls to load kexec kernel which only accept kernels signed with a certified keys or with no key. That would be a problem for testing test kernels which are usually signed with a test key.

Without this change, you will this error on fedora on a test kernel:
\# kdumpctl restart
kexec: unloaded kdump kernel
Stopping kdump: [OK]
Using kexec file based syscall.
kexec_file_load failed: Key was rejected by service
kexec: failed to load kdump kernel
kexec_file_load() failed, please try kexec_load()
Starting kdump: [FAILED]


kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+    elif [ "$ACTION" = "replace" ] && [ -z "$VALUE2" ]; then
+        Error "- Missing new_value for replacing."
+        return 1
+    elif ! grep "^$KEY=\"" "${KDUMP_SYS_CONFIG}"; then
+        Error "- Invalid KEY: $KEY."
+        return 1
+    fi
+
+    local kdump_sys_config_tmp="${KDUMP_SYS_CONFIG}.tmp"
+    \cp "${KDUMP_SYS_CONFIG}" "${kdump_sys_config_tmp}"
```
emanchado, line 534: What does the `\` mean, here?

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
         shift
     done
 
-    rstrnt-report-log -l "${KDUMP_CONFIG}"
+    RhtsSubmit "${KDUMP_CONFIG}"
```
emanchado, line 509: I thought we were trying to move away from the `Rhts*` function and towards using restraint?

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+        return 1
+    fi
+
+    local kdump_sys_config_tmp="${KDUMP_SYS_CONFIG}.tmp"
+    \cp "${KDUMP_SYS_CONFIG}" "${kdump_sys_config_tmp}"
+
+    # Note: When KEY is "KDUMP_COMMANDLINE", ACITION add/remove/replace is actually
+    # manipulating the value of current kernel cmdline and assign
+    # it to KDUMP_COMMANDLINE.
+    # if there is no value set to KDUMP_COMMANLINE, assign current kernel cmdline
```
emanchado, line 539: *`KDUMP_COMMANDLINE`

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+    # to it for later string manipluation.
+    if [ "$KEY" == "KDUMP_COMMANDLINE" ] && \
+        grep "^$KEY=\"[[:space:]]*\"" "${kdump_sys_config_tmp}"; then
```
emanchado, line 542: Should that grep regular expression have a `$` at the end? Or at least turn the `[[:space:]]*` into `[[:space:]]+`?

Otherwise it just checks that the `$KEY` is there, but not that it's empty.

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+
+        sed -i "/^KDUMP_COMMANDLINE=\"/d" "${kdump_sys_config_tmp}"
+        echo "KDUMP_COMMANDLINE=\"$(cat /proc/cmdline)\"" >> "${kdump_sys_config_tmp}"
```
emanchado, line 545: Why not replace with `sed` in one step, instead of removing the line first and then add it at the end?

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+
+    case $ACTION in
+        add)
+            # Check if the key already has a value on it, then append " $VALUE1" to the
+            # original value.
+            if grep "^$KEY=\"[[:space:]]*\"" "${kdump_sys_config_tmp}"; then
+                sed -i /^"$KEY="/d "${kdump_sys_config_tmp}"
+                echo "$KEY=\"$VALUE1\"" >> "${kdump_sys_config_tmp}"
+            else
+                sed -i  /^"$KEY="/s/\"/" $VALUE1"\"/2 "${kdump_sys_config_tmp}"
```
emanchado, line 556: I'm not sure I understand this line, but wouldn't it be more readable like this?
```
sed -i  "/^$KEY=/s/\"/ $VALUE1\"/2" "${kdump_sys_config_tmp}"
```
Also, wouldn't that replace the **first** double quote, instead of the last? Shouldn't the regular expression be `\"$`, with the dollar at the end?

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+
+    [ $? -ne 0 ] && {
+        Error "- Failed to edit kdump sysconfig"
+        rm -f "${kdump_sys_config_tmp}"
+        return 1
+    }
+
+    \mv "${kdump_sys_config_tmp}" "${KDUMP_SYS_CONFIG}"
+    RhtsSubmit "${KDUMP_SYS_CONFIG}"
+    sync;sync;sync
```
emanchado, line 583: Why three times??

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+    esac
+
+    [ $? -ne 0 ] && {
+        Error "- Failed to edit kdump sysconfig"
+        rm -f "${kdump_sys_config_tmp}"
+        return 1
+    }
+
+    \mv "${kdump_sys_config_tmp}" "${KDUMP_SYS_CONFIG}"
+    RhtsSubmit "${KDUMP_SYS_CONFIG}"
```
emanchado, line 582: Same comment here about the `Rhts*` -> restraint

kdump/include/runtest.sh
```diff
@@ -511,7 +506,82 @@ AppendConfig()
...
+        return 1
+    elif ! grep "^$KEY=\"" "${KDUMP_SYS_CONFIG}"; then
+        Error "- Invalid KEY: $KEY."
+        return 1
+    fi
+
+    local kdump_sys_config_tmp="${KDUMP_SYS_CONFIG}.tmp"
+    \cp "${KDUMP_SYS_CONFIG}" "${kdump_sys_config_tmp}"
+
+    # Note: When KEY is "KDUMP_COMMANDLINE", ACITION add/remove/replace is actually
```
emanchado, line 536: *`ACTION`

kdump/include/runtest.sh
emmawu: This is to overwrite the alias in case "cp -i" is set. Copied it from the original kdump test. May not be necessary in most cases, but no harm either.

kdump/include/runtest.sh
emmawu: If that is the case, put all rhts- cmd in a wrapper could make it easier to move all rhts-* cmd away in the future. Then what we need to do is to change the call inside RhtsSubmit, instead updating rhts- all over the place.

kdump/include/runtest.sh
emmawu: Thanks. Fixed.

kdump/include/runtest.sh
emmawu: Fixed. There are usually only 2 quotes in KDUMP_COMMANDLINE, as kernel options will normally not be provided with quotes. So the original idea is to check $KDUMP_COMMANDLINE="xxx" instead of KDUMP_COMMANDLINE="xxx"$

kdump/include/runtest.sh
emmawu: sed replace in one stop may look good. But it's more straight forwad to simply delete the old line and add the correct line back.

kdump/include/runtest.sh
emmawu: The original idea is to replace the second (which usually the last quote) as kernel options are normally provided without extra quotes. But you are right, \"$ would be more accurate. Fixed.

kdump/include/runtest.sh
emmawu: It happened some times that panic happened too sudden that not all test data are written to disk, and caused unexpected error. There is  an saying that sync; sync; sync would do the trick.

kdump/include/runtest.sh
emmawu: The rhts-* can be moved away all together in the furture within RhtsSubmit wrapper.

kdump/include/runtest.sh
emmawu: Thanks. Fixed.

emmawu:
Fixed 4 places as suggested. Pls review again.

kdump/include/runtest.sh
rasibley: @emmawu  thanks for the PR! Yes we have moved away from RHTS, by converting all the CKI tests to use Restraint compatible commands and removing the RHTS libraries sourced in the code. Would you mind just renaming the function to RstrntSubmit instead to avoid confusion ? Or feel free to file a separate PR. Thanks!

emmawu: Sure. Updated.

emmawu:
Renamed RhtsSubmit to RstrntSubmit. Pls review.

emmawu:
Improved the sed cmd a bit in AppendSysconfig. Tested passed in Beaker.
Please let me know if any other change need to be made.

kdump/include/runtest.sh
emanchado: Ah, nice, thanks! I never knew.

emanchado:
Updating the branch now. Should we run any tests after it's updated, before merging? In any case, it's fine by me, just merge if you want.

emmawu:
Ah, I accidently reverted my master with upstream and this PR is closed as a result :( I shouldn't file a PR from my master.
Sorry. I will create a new PR soon..

PR #524, opened 2020-05-12
kdump: Add test restore/cleanup and set KDUMP_FILE_LOAD to off

emmawu:
A few changes

- Added steps to restore default kdump config/systconfig and cleanup vmcore.
- Renamed RhtsSubmit to RstrntSubmit
- Do not uppload kdump initramfs img by default. It's for debugging purpose only.
- Set KDUMP_FIlE_LOAD to off. In Fedora, there is a feature change made in kdump that by default it will use kexec_load_file calls to load kexec kernel which only accept kernels signed with a certified keys or with no key. That would be a problem for testing test kernels which are usually signed with a test key.

Without this change, you will this error on fedora on a test kernel:
```
# kdumpctl restart
kexec: unloaded kdump kernel
Stopping kdump: [OK]
Using kexec file based syscall.
kexec_file_load failed: Key was rejected by service
kexec: failed to load kdump kernel
kexec_file_load() failed, please try kexec_load()
Starting kdump: [FAILED]
```


emmawu:
It's same as https://github.com/CKI-project/tests-beaker/pull/522.
Please review.

emmawu:
Pushed a tiny fix for UPLOADRD option.

Btw, when I tested this PR, it got this "uploading.. Conflict" error twice (All other rstrnt-report-log -l worked fine):  
Error uploading: /mnt/testarea/crash-simple.vmcore.log Conflict [soup_http_error_quark, 409]

It's unrelated to this PR. But do you have any idea why this restraint report cmd failed?

kdump/include/runtest.sh
```diff
@@ -511,7 +506,86 @@ AppendConfig()
...
+
+    [ $? -ne 0 ] && {
+        Error "- Failed to edit kdump sysconfig"
+        rm -f "${kdump_sys_config_tmp}"
+        return 1
+    }
+
+    \mv "${kdump_sys_config_tmp}" "${KDUMP_SYS_CONFIG}"
+    RstrntSubmit "${KDUMP_SYS_CONFIG}"
+    sync;sync;sync
```
emanchado, line 587: Now that I think of it, maybe add a comment with the explanation why three `sync` calls?

kdump/include/runtest.sh
```diff
@@ -579,9 +654,11 @@ RestartKdump()
     if grep -v -E "$skip_pat" /tmp/kdump_restart.log |  grep -i -E "can't|error|warn";  then
         Warn 'Restarting kdump reported warn/error message'
     fi
-
     sync;
-    ReportKdumprd
+
+    if [ "$UPLOADRD" == true ]; then
```
emanchado, line 659: Nitpick: isn't the convention to use `"${UPLOADRD}"`?

kdump/include/runtest.sh
emmawu: Updated. Added a comment next to 3 sync calls.

kdump/include/runtest.sh
emmawu: updated.

emmawu:
Fixed "${UPLOADRD}" and one comment for 3 sync.

rasibley:
Thanks @emmawu I haven't seen this error, can you send me the beaker job which contains it ? I did a tet job myself and it ran perfectly: see J:4250486

PR #525, opened 2020-05-12
networking: do not run netperf as netperf fail to compile

linup2011:


rasibley:
I have the same question, isn't netperf needed for several networking cases ? It's hopefully just a simple change to to get it work with gcc10, see example patch and doc about -fno-common: 
https://gcc.gnu.org/gcc-10/porting_to.html 
https://github.com/jstancek/amtu/commit/f304cf2a1a6e31048086884ffb1b8644e63f40b9

linup2011:
for the tier1 test, I think netperf is not mandatory

linup2011:
> I have the same question, isn't netperf needed for several networking cases ? It's hopefully just a simple change to to get it work with gcc10, see example patch and doc about -fno-common:
> https://gcc.gnu.org/gcc-10/porting_to.html

the solution seems to be simple, I'll try to fix netperf code if we need to use netperf anyway
> [jstancek/amtu@f304cf2](https://github.com/jstancek/amtu/commit/f304cf2a1a6e31048086884ffb1b8644e63f40b9)

rasibley:
Thanks, if it's not needed, I agree we should remove it, @cki-bot please test.

rasibley:
@emanchado I'm on PTO the next few days but feel free to merge if results look good from the bot :-) Then we'll need to remember to re-enable this test (call the bot to test just upstream-stable) for upstream/ark in kpet-db. https://gitlab.com/rasibley/kpet-db/-/blob/master/suites/networking/vnic/ipvlan/basic/index.yaml#L22

emanchado:
Again that pesky fanotify failure, and the rest seem harmless/random. Merging.

@rasibley: To reenable that test you talk about, do you mean we can do it as soon as this is merged, or do we have to wait for anything else?

PR #526, opened 2020-05-12
af_alg-test: Skip multithreaded tests

WOnder93:
There is a race condition when these are run with the long XTS test
vectors, so skip them for now.


rasibley (approved):
Thanks, LGTM !

PR #528, opened 2020-05-13
Kselftest update

liuhangbin:
This patchset fixed selftest URL typo and skip the test if not find.
I also fixed a parameter issue in include file.

### DIFF
--- a/kselftests/Makefile
+++ b/kselftests/Makefile
@@ -78,7 +78,7 @@ $(METADATA): Makefile
 	@echo "RunFor:       $(PACKAGE_NAME)" >> $(METADATA)
 # add any other packages for which your test ought to run here
 	@echo "Requires:     tar autoconf automake" >> $(METADATA)
-	@echo "Requires:     iproute-tc jq netsniff-ng" >> $(METADATA)
+	@echo "Requires:     wget iproute-tc jq netsniff-ng" >> $(METADATA)
 	@echo "Requires:     clang valgrind" >> $(METADATA)
 	@echo "Requires:     $(PACKAGE_NAME)" >> $(METADATA)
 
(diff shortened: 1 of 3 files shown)


rasibley:
Thanks @liuhangbin  ! Can you please add wget to the Makefile ? Otherwise it will abort since it's missing the dependency to download the tarball. I tested out your changes in J:4252724, I had to set the env variable in the xml, we still need to set the task parameter in kpet-db somehow, I left a question for @spbnick in https://gitlab.com/cki-project/kpet-db/-/merge_requests/3

I see the non bpf subests are skipped, but the bpf tests have some failures, cc @veruu 

Also fyi, as you can see, we just published kpet-db, I moved your existing MR to gitlab.com :-)

veruu:
Thanks @liuhangbin the functionality is looking great!

@rasibley I think those failures are something to be resolved in bpf tree - it looks like it tests require some extra files that are built, but they are not added to the actual list of dependencies for selftests to package so they are not copied to the tarball. Still looking into that but this is what I think the problem is, it wouldn't be the first time we ran into this as bpf people are only running the tests locally :/ I'll talk about this with them later today. It shouldn't be a blocker to merge this PR once wget is added.

veruu:
@rasibley looks like one of our lovely bpf people already noticed the same problem and posted a patch yesterday: https://lore.kernel.org/bpf/20200513021722.7787-1-yauheni.kaliuta@redhat.com/

veruu:
@rasibley fyi I tested the new kernel with the above patch, it's causing a different error during "make install" and we'll have to wait for it to be resolved.

I see wget is added to this PR already and the bug mentioned above is not a blocker for this PR, so IMO we can merge :)

rasibley:
Thanks @veruu ok I think we can merge as well, and according to Nick all the magic happens in kpet to assign the correct tarball url. I'll see if I can trigger the bot there against bpf-next tree.

PR #529, opened 2020-05-13
loopdev/sanity: fix gcc10 build error

jencce:
With another commit to print more debug info for xfstests.


rasibley:
Thanks @jencce did you mean to update both xfstests and loopdev/sanity in the same PR, they seem unrelated ? Do you have a test job for loopdev/sanity on F32 ? Otherwise we can test it out when we re-enable in kpet-db.

jencce:
Sorry for the late response. I only tested the loopdev/sanity change manually. 
The xfstests debug message commit is useful but it's unrelated with the gcc10 issue. It got lost in my last pr, so I included it in this one. Sorry for the confusion I should have explained it explicitly.

rasibley:
No problem, thanks for clarifying @jencce !

PR #533, opened 2020-05-14
Add fanotify09 to knownissues

rasibley:
This test is failing as it's missing patches from mainline, mask in knownissue file until fixes make it to stable tree.


rasibley:
Never mind these are in stable now, so no need to mask it.

PR #536, opened 2020-05-17
Minor libkcapi test improvements

WOnder93:
Explanations are in the commit messages.


packages/libkcapi/af_alg-test/runtest.sh
```diff
@@ -29,8 +29,6 @@ rlJournalStart
         rlRun "git clone '$GIT_URL' libkcapi"
         rlRun "(cd libkcapi && git checkout $GIT_REF)"
         rlRun "(cd libkcapi && autoreconf -i)"
```
emanchado, line 31: The `autoreconf` line is still needed?

WOnder93: Yes, the `test-invocation.sh` script needs the `configure` script to be generated already.

emanchado:
Should I merge, or do you want a second review or something?

WOnder93:
You can merge it, as far as I'm concerned. Thanks for the review!

PR #538, opened 2020-05-20
kdump: Test restore shouldn't clear reboot flag

emmawu:
File $K_REBOOT was wrongly cleared in RestoreKdumpConfig().
It will cause system reboot loop when KER1ARGS passed to the test requiring an extra reboot for changing boot loader.


emmawu:
Tested on RHEL8

PR #539, opened 2020-05-22
ftrace-tracer: Limit the output of function and function_graph

sztsian:
In busy systems function and function_graph could have a lot of output
and when we cat the trace data the ring buffer increases faster than
cat, which results in infinite waiting. So now switch to show only the
top lines to reduce the chance of aborts in test.


trace/ftrace/tracer/ftrace-tracer.sh
```diff
@@ -69,7 +69,7 @@ tracername="function_graph"
 if grep -qw ${tracername} ${FTRACE_PREFIX}/available_tracers; then
   starttrace ${tracername}
   sleep 5
-  cat ${TRACE} &> /dev/null ; rtrn=$?
+  cat ${TRACE} | head &> /dev/null ; rtrn=$?
```
emanchado, line 72: Wait, this would actually be better as `head ${TRACE} &>/dev/null`. No need to use `cat` here if you're going to use `head` anyway.

sztsian: I'll try later on. I use cat previously to reduce the chance of infinite waiting (but it did not help enough). I'll see if head can do enough trick or not then follow-up with email or comment.

sztsian:
Taking all into consideration, I think it is easier to switch all `cat` into `head`, so this eventually reduced the chance of infinite waiting for all ftrace tracers. So I just did a rebase to implement that.

emanchado:
@cki-bot: please test this!

emanchado:
The `fanotify` failures are due to some missing kernel patches, apparently, so I'll merge.

PR #540, opened 2020-05-25
af_alg-test: Bump upstream ref to v1.2.0

WOnder93:
The latest release disables some tests that were broken. We can now
remove the workaround and this will also fix the false positives on
machines with async crypto drivers.

### DIFF
--- a/packages/libkcapi/af_alg-test/runtest.sh
+++ b/packages/libkcapi/af_alg-test/runtest.sh
@@ -22,7 +22,7 @@
 . /usr/share/beakerlib/beakerlib.sh
 
 GIT_URL="https://github.com/smuellerDD/libkcapi"
-GIT_REF="c77b0cf4567e9c534378d7c5bfeceb3eda1d2d30"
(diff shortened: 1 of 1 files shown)


emanchado:
@cki-bot: test this, please!

emanchado (approved):
LGTM. And nice commit message!

I'll wait until the tests finish before merging.

emanchado:
Hm, weird, there seems to be a consistent (two testruns in a row) failures in two `fanotify*` tests. Could that be related to this?

WOnder93:
How could it be related? The Libkcapi test runs after LTP...

veruu:
@emanchado the fanotify LTP failures are due to missing kernel patches. As @WOnder93 mentioned, it's a completely different test suite that's running before the test he's changing here, that is already a good indicator the failures can't be related to this PR ;)

rasibley (approved):
LGTM ! Thanks @WOnder93  !

PR #541, opened 2020-05-26
WIP: kpkginstall fixes

veruu:
see separate commits for explanation

### DIFF
--- a/distribution/kpkginstall/runtest.sh
+++ b/distribution/kpkginstall/runtest.sh
@@ -257,6 +257,10 @@ function targz_install()
         sed -i "s/title.*/$title/" "${f}"
         cki_print_success "Removed trailing whitespace in title record of $f"
       done
+
+      # Workaround for BZ 1698363
+      grubby --set-default /boot/vmlinuz-"${KVER}" && zipl
+      cki_print_success "Grubby workaround for s390x completed"
   fi
 }
 
--- a/distribution/kpkginstall/runtest.sh
+++ b/distribution/kpkginstall/runtest.sh
@@ -469,7 +469,7 @@ else
       "${KVER}+debug"           # RHEL 8 style debug kernels
     )
   else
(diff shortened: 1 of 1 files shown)


veruu:
I cloned Emma's job with my kpkginstall branch to test fixes, J:4278996 is currently running.

@cki-bot do an extra test to verify nothing broke please!

veruu:
I'm having a hard time getting FC32 installed on s390x, marking as WIP in the meanwhile until I get an actual test job.

veruu:
The test results for RHEL and non-s390x upstream jobs are looking good so no unexpected breakages. There have been some aborts unrelated to the changes.

I'm still working on getting an upstream s390x job.

veruu:
Random upstream (not the "similar NVR" one) test run looks ok, see R:8287322

I cancelled the other s390x recipes submitted by bot (that didn't run yet) to save on resources.

Hopefully R:8288092 will provision

rasibley:
Changes LGTM, and the bot results look good, for the upstream jobs (including s390x) I do see the following error in the log but it didn't cause it to fail, might have been there previously:
` ./runtest.sh: line 165: /tmp/kpkginstall/KPKG_KVER: No such file or directory`

See J:4279020

veruu:
Yes, that error is there all the time. Most likely since Major wrote the code that assumed /tmp stays alive after reboot :facepalm: It's not causing any problems AFAIK

rasibley:
LOL ok well I think it's safe to merge since the bot was able to test it on s390x/upstream, are we good to merge then or are you waiting on the other job to finish up ?

veruu:
I'd like to verify the bug we saw is actually fixed, but at the same time since bot results were good it should be safe to merge anyways. So if you are ok with the changes we can merge and follow up later if it's needed

rasibley:
Sounds good to me, I'll go ahead and merge and we can follow up later if needed.

PR #542, opened 2020-05-26
Update LTP to recent commit

rasibley:
* Update LTP to a recent commit to pull in hugetlbfs fixes
* Add bpf/s390x failures to known issues file
* Miscellaneous cleanup for known issues file
* Remove pty04, move_pages12, and connect02 from known issues

### DIFF
--- a/distribution/ltp-upstream/include/knownissue.sh
+++ b/distribution/ltp-upstream/include/knownissue.sh
@@ -121,15 +121,11 @@ function knownissue_filter()
 	[ $(free -g | grep "^Mem:" | awk '{print $2}') -gt 8 ] && tskip "oom0.*" fatal
 	# copy_file_range02 is a new unstable test case and changes too frequently
 	tskip "copy_file_range02" unfix 
-	# move_pages12 patches pending
-	# http://lists.linux.it/pipermail/ltp/2019-July/012907.html  --- patch pending on review in upstream
-	# http://lists.linux.it/pipermail/ltp/2019-July/012962.html  --- kernel bug in upstream kernel
-	tskip "move_pages12" unfix
 	# Issue TBD
 	tskip "madvise09" fatal
 	# https://github.com/linux-test-project/ltp/issues/611
-	tskip "ksm03.*" fatal
-	tskip "ksm04.*" fatal
+	tskip "ksm03.* ksm04.*" fatal
+	tskip "ksm02.*" unfix
 	# Bug 1660161 - [RHEL8] ltp/generic commands mkswap01 fails to create by-UUID device node in aarch64
 	# Issue TBD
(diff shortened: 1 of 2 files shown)


rasibley:
Test job: 4280416 @veruu can you review please ?

veruu:
Changes and results look good to me. However, the test job is running with an older kernel where the BPF loading issues don't manifest, so we can't actually verify that we're skipping all the tests we need.

Should we wait for the LTP maintainer ack before merging?

rasibley:
@veruu good point! I'll retest with the right kernel, I had just grabbed the latest from CKI jobs not paying attention to that detail, plus Jan pointed out some tests I can remove from known issues. Will include that in the next round.

rasibley:
Test Job with the right kernel J:4283650 and maybe @jstancek or @wangli5665 can give an ack while we wait on test results ?

rasibley:
Results are all green except one new failure on aarch64: 
```
>>> 12:hugeshmat04 BROK <<<
###############################################################################
# Test Num    : 12                                                            #
# Test Case   : hugeshmat04                                                   #
# Test Result : BROK                                                          #
###############################################################################

     1	<<<test_start>>>
     2	tag=hugeshmat04 stime=1590603615
     3	cmdline="hugeshmat04 -i 5"
     4	contacts=""
     5	analysis=exit
     6	<<<test_output>>>
     7	tst_hugepage.c:50: INFO: 1 hugepage(s) reserved
     8	tst_test.c:1244: INFO: Timeout per run is 0h 05m 00s
     9	tst_hugepage.c:50: INFO: 513 hugepage(s) reserved
    10	hugeshmat04.c:60: BROK: shmat: EINVAL (22)
```

wangli5665:
> on aarch64:

>      9	tst_hugepage.c:50: INFO: 513 hugepage(s) reserved
>     10	hugeshmat04.c:60: BROK: shmat: EINVAL (22)

This failure looks strange to me, hugeshmat04 has already reserved hpages(513) and allocate shared memory SIZE(512 hpaegs) success, but get an error in attach the shmid to the address spaces.

From the manual page:
    "EINVAL     Invalid shmid value, unaligned (i.e., not page-aligned and SHM_RND was not specified) or invalid shmaddr value, or can't attach segment at shmaddr, or SHM_REMAP was specified and shmaddr was NULL."

It seems not caused by LTP test, I would reserve this machine to have a close debug.

@rasibley In a word, it does not matter to update ltp commit I think, we can mark this fail and add in blacklist if possible.

rasibley:
Thanks @wangli5665, I can see if I can reproduce and add to the known issues in a separate PR, but I agree unrelated. @veruu ok to merge ?

veruu:
Yes, thank you!

PR #546, opened 2020-05-29
kdump: Pull latest kdump/crash from F33

emmawu:
Since base compose now is Fedora-32, pull latest kdump/crash packges from F-33.


emmawu:
I've tested it on Beaker [J:4290004](https://beaker.engineering.redhat.com/jobs/4290004) on F32 x86_64. And if the test does fail on upstream kernel, it could be a true bug...

emanchado:
Ah, excellent, thanks! I'll merge, then.

